PR #93, opened 2014-03-08
Fix for readline add_history with interactive mode

rfdrake:
If you use interactive mode, augtool never adds anything to history that
you type once the input becomes a normal tty.

This is because it was checking stdin, which as far as the system is
concerned is still !isatty().  I'm not sure of the better way to fix
this.  An freopen stdin with /dev/tty might be good, but just changing
the add_history check to use rl_instream seems to work fine as long as
it doesn't cause unintended effects.

### DIFF
--- a/src/augtool.c
+++ b/src/augtool.c
@@ -404,7 +404,7 @@ static int run_command(const char *line) {
     int result;
 
     result = aug_srun(aug, stdout, line);
-    if (isatty(fileno(rl_instream)))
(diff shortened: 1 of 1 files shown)


lutter:
Thanks for the bug fix. I made two minor changes:

* squashed the two commits into one
* use freopen instead of expl;icit fclose/fopen

I am very sorry it took this long to get this in ...

Committed as acfa635f

Issue #88, opened 2014-02-24
Undefined refernce to YYID


levsh:
Hi!
build augeas failed for me
Ubuntu 12.04 64bit,  gcc version 4.6.3 (Ubuntu/Linaro 4.6.3-1ubuntu5), bison (GNU Bison) 3.0.2 from sources.
Error:

``` bash
  CCLD   liblexer.la
  CCLD   libaugeas.la
  CC     augtool.o
  CCLD   augtool
./.libs/libaugeas.so: undefined reference to `YYID'
collect2: ld returned 1 exit status
make[3]: *** [augtool] Error 1
make[3]: Leaving directory `/tmp/augeas/src'
make[2]: *** [all] Error 2
make[2]: Leaving directory `/tmp/augeas/src'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/tmp/augeas'
make: *** [all] Error 2

lutter:
This was fixed in 3408111d

PR #111, opened 2014-04-09
Fix ubuntu bug 1302638

JurjenBokma:
Opening this PR as requested by Raphaël Pinson in https://bugs.launchpad.net/ubuntu/+source/augeas/+bug/1302638
This is my first time doing that, so if I'm doing it wrong, please correct me.

When building, at the linking stage I run into this issue: https://github.com/hercules-team/augeas/issues/88
My topic branch `fix-YYID_undefined_reference' trivially works around that. I should 've bisected the tree, but haven't the time. I suspect it may actually be an autotools issue?

This fix-ubuntu-bug-1302638 makes the krb5.aug lens process lower case realms in the [realms] section, but not in the [appdefaults]. So there are still valid krb5.conf that the lense will not parse. This seems to be by design: in order to be able to label an entry 'app' or 'realm' in the appdefaults section.
I see only ways out of this situation that break backward compatibility.

Thanks for reviewing!
Jurjen

### DIFF
--- a/lenses/krb5.aug
+++ b/lenses/krb5.aug
@@ -22,6 +22,7 @@ let closebr = del /[ \t]*\}/ "}"
 *)
 
 let realm_re = /[A-Z][.a-zA-Z0-9-]*/
+let realm_anycase_re = /[A-Za-z][.a-zA-Z0-9-]*/
 let app_re = /[a-z][a-zA-Z0-9_]*/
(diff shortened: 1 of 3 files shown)


lutter:
I reworked this a bit and merged (spiritually similar) patches 3408111d, 7db96415 and db86863c

Thanks for the contribution and for your patience ...

PR #112, opened 2014-04-10
Fix ubuntu bug 1302638

JurjenBokma:
Opened this PR as per Raphaël Pinson's request, and after accidentally closing another, identical PR.
It is intended to fix https://bugs.launchpad.net/ubuntu/+source/augeas/+bug/1302638

With this fix applied, the parse tree is unchanged, but the krb5 lens also parses lower case realms in the [realms] section, though not in the appdefaults section. If my judgement is correct, it is impossible to make the lens parse all input that is valid to the Kerberos library without causing backward incompatibility.
That is because the lens tries to distinguish (by case) between application names and realm names in the [appdefaults] section, which is something the library doesn't do.

The undefined YYID reference I encountered is unrelated to that bug, but I had to fix it to get augeas to build. It may actually be an autotools problem?

Thank you for reviewing!

### DIFF
--- a/lenses/krb5.aug
+++ b/lenses/krb5.aug
@@ -22,6 +22,7 @@ let closebr = del /[ \t]*\}/ "}"
 *)
 
 let realm_re = /[A-Z][.a-zA-Z0-9-]*/
+let realm_anycase_re = /[A-Za-z][.a-zA-Z0-9-]*/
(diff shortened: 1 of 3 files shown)


lutter:
This duplicates PR 111 which I just merged

PR #135, opened 2014-06-13
Add aug_filter_file, cmd_filter_file and --filter-file

raphink:
This PR adds a new `-F|--filter-file` option to `augtool` which allows to choose which lens to load based on the desired files. It loads `augtool` with `--noload` and filters the lenses dynamically.

The result is faster than loading all lenses, but still definitely slower than loading with a specific lens (since it still needs to load all the lenses in order to determine which ones to keep).

Sample usage:

    $ augtool -F /etc/fstab -F /etc/hosts
    augtool> ls /files/etc/
    mtab/ = (none)
    fstab/ = (none)
    hosts/ = (none)

The PR exposes a new public API call `aug_filter_file`:

```c
/*
 * Function: aug_filter_file
 *
 * Load a specific FILE, using autoload statements.
 * You need to call aug_load after this action.
 *
 * Returns:
 * 1 on success, -1 on failure
 */
int aug_filter_file(augeas *aug, const char *file);
```

It also exposes a `filter-file` command to add more mappings:

```
augtool> help filter-file
  COMMAND
    filter-file - load a specific file

  SYNOPSIS
    filter-file <FILE>

  DESCRIPTION
    Load a specific FILE, using autoload statements.
   You need to call load after this action.
   

  OPTIONS
     <FILE>    the file to filter

```


raphink:
@lutter ping?

raphink:
@lutter is this something you might still want in?

lutter:
Yes, I think it would be a good addition. I am still somewhat hung up on how the code currently remembers that a filter was added and therefore subsequent attempts to add files should not clear out `/augeas/load`; basically, I think that function does too much, and is too opinionated about how callers want to manage the tree.

For the API, it would be better to change `aug_filter_file` to `aug_add_filter` or similar, i.e., make that API call additive; or provide a call `aug_lens_for_file(augeas *aug, const char *path, char **lens)` that looks up the lens that would autoload that file. Callers can then decide themselves what they want to do with that information; e.g., the could clear out `/augeas/load` and call `aug_transform` repeatedly. I would implement `aug_lens_for_file` by looking at `aug->modules` rather than depending on anything in the tree. The heart of that would be something like `transform_applies` in `transform.c`, except with a slightly different method signature `int transform_applies_struct(struct transform *, const char *path)` (and obviously, a better name)

For `augtool`, would make `-F` work the same as `-t`, except that you don't have to specify the lens; i.e., running `augtool -F /etc/fstab` would be the same as `augtool -t Fstab.lns incl /etc/fstab`.

Since this is probably a bit more involved than what you had in mind, I'd be more than happy to help, especially with `aug_lens_for_file`.

If you think it's needed, we could also add `aug_add_filter`, which would basically call `aug_lens_for_file` and then `aug_transform`

raphink:
Thank you for your reply.

I was not aware of `aug->modules`, so I'll have a look into that. It makes sense to try to avoid manipulating the tree as much as possible.

> For `augtool`, would make `-F` work the same as `-t`, except that you don't have to specify the lens; i.e., running `augtool -F /etc/fstab` would be the same as `augtool -t Fstab.lns incl /etc/fstab`.


There is a difference though. `-F` only makes sense with `-A`, because otherwise all known files are autoloaded anyway, whereas `-t` still makes sense without `-A` since you're manually associating a path with a lens.

lutter:
> There is a difference though. -F only makes sense with -A, because otherwise all known files are autoloaded anyway, whereas -t still makes sense without -A since you're manually associating a path with a lens.

Good point. We should make -F imply -A then, right ?

raphink:
Yes, I think so.

raphink:
Just rebased the current branch for a start.

raphink:
@lutter it seems to me that autoload infos requires compiling (in `load_module_file`), right?

raphink:
The latest commit takes a new approach, using `load_file` and `aug->modules`.

lutter:
> it seems to me that autoload infos requires compiling (in load_module_file), right?

Yes, `aug->modules` gets set up by `interpreter_init` which only loads modules when `-A` is _not_ given - I didn't think about that we need modules loaded, but `/augeas/load` not populated for `-F`. I am not entirely sure what the right answer here is, but it seems we need to add a flag to `aug_flags` that has the effect of loading modules but not populating `/augeas/load`. We wouldn't need to expose that flag in `augtool` as we'd only need it when the user also specifies `-F`, I think. We would then have these three flags to control initialization:

* `AUG_NO_MODL_AUTOLOAD`: don't load modules, do not populate `/augeas/load`, do not load files
* `AUG_NO_XFM`: load modules, do not populate `/augeas/load`, do not load files
* `AUG_NO_LOAD`: load modules, populate `/augeas/load`, do not load files

I don't particularly like `AUG_NO_XFM` for the flag, but can't think of a better one right now ...

lutter:
Actually, rather than a new flag for `aug_flags`, couldn't we just be a little wasteful and have `augtool` set `AUG_NO_LOAD` and then `rm /augeas/load` right after it called `aug_init` ?

raphink:
@lutter The approach I've taken in my new implementation is to use `AUG_NO_LOAD` indeed (or `-L` in `augtool`):

```
src/augtool -I lenses/ -L 
augtool> ls /files
augtool> filter-file "/etc/fstab"
augtool> ls /files/etc/
fstab/ = (none)
```

or using `-F`, implying a `-L`:

```
src/augtool -I lenses/ -F /etc/fstab -F /etc/passwd
augtool> ls /files/etc/
fstab/ = (none)
passwd/ = (none)
```

I think that should be clear enough. `/augeas/load` *is* fully populated, but `aug_load` is not called, so it's just like using `-L` and cherry-picking which files you want to load, and you can call `filter-file` as many times as you want to load more specific files.

thedrow:
This will optimize the case when we ask about one file but don't know which file it is in advance.
I'm going to use it for https://github.com/facebook/osquery/pull/2181 if the WHERE clause requires only a specific file/files.

raphink:
@lutter what do you think of the new implementation?

lutter:
I really like the direction of the latest iteration of this; basically a targeted way to load individual files. Maybe we should call the API call `aug_load_file` and the command line option `--load-file|-l`. I'll leave some more detailed comments on the changes themselves. For the next round of review, it would be good to squash/reorganize the patches to make them a bit more readable (there's now a lot of stuff that patches introduce that later gets undone by other patches)

src/augeas.c
```diff
@@ -83,7 +83,8 @@ static const char *const errcodes[] = {
     "Failed to execute command",                        /* AUG_ECMDRUN */
     "Invalid argument in function call",                /* AUG_EBADARG */
     "Invalid label",                                    /* AUG_ELABEL */
-    "Cannot copy node into its descendant"              /* AUG_ECPDESC */
+    "Cannot copy node into its descendant",             /* AUG_ECPDESC */
+    "Cannot setup file filter"                          /* AUG_ESFILTER */
```
lutter: The way this is used, wouldn't `Cannot find a lens for the given file` be more accurate ?

src/transform.h
```diff
@@ -68,6 +68,8 @@ int transform_validate(struct augeas *aug, struct tree *xfm);
  */
 int transform_load(struct augeas *aug, struct tree *xfm);
 
+int load_file(struct augeas *aug, struct lens *lens, const char *lens_name, char *filename);
```
lutter: I think `load_file` is a little too low-level here.  It might be better to change the signature of `transform_load` to
```
int transform_load(struct augeas * struct tree *xfm, const char *file)
```
and change the implementation such that `transform_load` behaves the way it behaves today when you pass `NULL` for `file`, and if `file` is anything else, only actually load that file.

src/augeas.c
```diff
@@ -2020,6 +2021,39 @@ int aug_escape_name(augeas *aug, const char *in, char **out) {
...
 }
 
+int aug_filter_file(struct augeas *aug, const char *file) {
+    int result = -1;
+
+    api_entry(aug);
+
+    bool found = false;
+
+    list_for_each(module, aug->modules) {
```
lutter: With the changes to `transform_load` that I suggested, and since you now work with the transforms set up in `/augeas/load`, this loop could become something like
```
/* Setup basically copied from aug_load */
struct tree *meta = tree_child_cr(aug->origin, s_augeas);
struct tree *load = tree_child_cr(meta, s_load);

list_for_each(xfm, load->children)  {
  if (filter_matches(xfm, file)) { /* Needs to be added to transform.h */
    transform_load(aug, xfm, file);
    break;
  } 
}
```

src/augeas.c
```diff
@@ -2020,6 +2021,39 @@ int aug_escape_name(augeas *aug, const char *in, char **out) {
     return result;
 }
 
+int aug_filter_file(struct augeas *aug, const char *file) {
```
lutter: Since this now actually loads the file, I think it should be renamed to `aug_load_file` (and similar for the augtool command and command line option)

raphink:
Thanks for all the comments @lutter . Here's the squash already, to have a clean base to work on.

src/transform.h
raphink: Better yet, is there a way I can generate an `xfm` to pass to `transform_load()`? This would allow to even pass globs.

src/augeas.c
raphink: Oh I see now. If I pass an `xfm` without a `file`, then it will load all the files for that `xfm`. Makes sense…

raphink:
@lutter implemented, and rebased again.

src/augeas.c
```diff
@@ -83,7 +83,8 @@ static const char *const errcodes[] = {
     "Failed to execute command",                        /* AUG_ECMDRUN */
     "Invalid argument in function call",                /* AUG_EBADARG */
     "Invalid label",                                    /* AUG_ELABEL */
-    "Cannot copy node into its descendant"              /* AUG_ECPDESC */
+    "Cannot copy node into its descendant",             /* AUG_ECPDESC */
+    "Cannot find a lens for file"                       /* AUG_ESLOADFILE */
```
lutter: Rather than create a new error code, I think we can use `AUG_ENOLENS` here.

src/augeas.c
```diff
@@ -2020,6 +2021,35 @@ int aug_escape_name(augeas *aug, const char *in, char **out) {
...
+        found = true;
+        break;
+      } 
+    }
+
+    ERR_THROW(!found, aug, AUG_ESLOADFILE, "can not determine lens to load file %s", file);
+
+    api_exit(aug);
+    result = 0;
+    return result;
```
lutter: These three lines should just become `result = 0` and fall through to the `error` label so that any cleanup that might be added in the future happens both when we return success and failure.

src/augeas.c
```diff
@@ -2020,6 +2021,35 @@ int aug_escape_name(augeas *aug, const char *in, char **out) {
...
+    struct tree *load = tree_child_cr(meta, s_load);
+    bool found = false;
+
+    api_entry(aug);
+
+    ERR_NOMEM(load == NULL, aug);
+
+    list_for_each(xfm, load->children)  {
+      if (filter_matches(xfm, file)) {
+        transform_load(aug, xfm, file);
```
lutter: One thing I noticed in my testing is that the tree thet gets created is marked dirty; issuing `save` right after loading a file will try to save the file, which is not correct. I'll provide a patch that shows how to call `tree_clean` to mark the subtree as clean and therefore avoid that.

src/augrun.c
```diff
@@ -1010,6 +1010,33 @@ static const struct command_def cmd_transform_def = {
     .help = cmd_transform_help
 };
 
+static void cmd_load_file(struct command *cmd) {
+    const char *file = arg_value(cmd, "file");
+    int r = 0;
+
+    r = aug_load_file(cmd->aug, file);
+    if (r == -1)
```
lutter: It's more robust to check `r < 0` here (in case we ever use other negative numbers to indicate other error conditions)

src/augtool.c
```diff
@@ -291,27 +293,28 @@ static void help(void) {
...
+    fprintf(stderr, "  -f, --file FILE        read commands from FILE\n");
+    fprintf(stderr, "  -s, --autosave         automatically save at the end of instructions\n");
+    fprintf(stderr, "  -i, --interactive      run an interactive shell after evaluating\n"
+                    "                         the commands in STDIN and FILE\n");
+    fprintf(stderr, "  -S, --nostdinc         do not search the builtin default directories\n"
+                    "                         for modules\n");
+    fprintf(stderr, "  -L, --noload           do not load any files into the tree on startup\n");
+    fprintf(stderr, "  -A, --noautoload       do not autoload modules from the search path\n");
+    fprintf(stderr, "  --span                 load span positions for nodes related to a file\n");
     fprintf(stderr, "  --timing             after executing each command, show how long it took\n");
```
lutter: The helptext needs to be shifted over here, too

src/augtool.c
```diff
@@ -649,6 +658,24 @@ static void add_transforms(char *ts, size_t tslen) {
...
+    char *command;
+    int r;
+    char *t = NULL;
+
+    while ((t = argz_next(ts, tslen, t))) {
+        r = xasprintf(&command, "load-file %s", t);
+        if (r < 0)
+            fprintf(stderr, "error: Failed to load file %s: could not allocate memory\n", t);
+
+        r = aug_srun(aug, stdout, command);
```
lutter: Just out of curiosity: why does this go through `aug_srun` instead of calling `aug_load_file` directly ?

src/transform.c
```diff
@@ -822,12 +822,17 @@ int transform_load(struct augeas *aug, struct tree *xfm) {
...
     }
+
     r = filter_generate(xfm, aug->root, &nmatches, &matches);
     if (r == -1)
         return -1;
     for (int i=0; i < nmatches; i++) {
         const char *filename = matches[i] + strlen(aug->root) - 1;
         struct tree *finfo = file_info(aug, filename);
+
+        if (file != NULL && strcmp(filename, file) != 0)
```
lutter: Instead of `strcmp` use `STREQ(filename, file)` here for better readability

lutter:
Overall, besides the minor comments above, this looks great ! I put some more suggestions up as https://github.com/lutter/augeas/commit/efff7c143992feb1039f040ffa87af744f40d961 on my branch. Would be great if you could work them in.

Besides addressing the comments, the only other thing that is missing are some tests. At a minimum, we need tests for `aug_load_file` in `tests/test-api.c` (could be just one more test function there) that check
* successfully loading a file
* getting `AUG_ENOLENS` when trying to load a file that has no lens associated
* `aug_load_file` succeeds when trying to load a nonexistent file (and does not put anything into the tree)

Once that is done, I'd like to have one more quick look, but I think we'd then be good to merge this. Great work !

src/augeas.c
raphink: How can you tell if the tree is dirty?

src/augtool.c
raphink: I guess I had implemented it this way 2 years ago for reason I can't remember.

src/transform.c
raphink: Ack

src/augtool.c
raphink: No actually, it's because `aug_srun` was written as an abstraction for augtool, so augtool commands are supposed to rely on it iirc.

raphink:
@lutter I fixed most comments and squashed again.

src/augeas.c
lutter: Each tree node has a `dirty` flag; whenever a tree node is modified, its `dirty` flag is set, as well as the `dirt` flags of all its ancestors. That is done by `tree_mark_dirty` in `augeas.c`. `tree_clean` does the reverse: clearing the flag on a node and all its descendants. `aug_save` (and `tree_save`, its internal cousin) only tries to save files that are marked as dirty to avoid lots of unnecessary tree -> file transformations.

src/augtool.c
lutter: Ok. I don't have a strong opinion either way; it's totally fine, I was just curious.

raphink:
@lutter can I just merge and squash https://github.com/lutter/augeas/commit/efff7c143992feb1039f040ffa87af744f40d961?

lutter:
Into your PR ? Yes, absolutely. All that's missing to merge into augeas are the tests, I think.

raphink:
OK. Cherry-picked and squashed. I'll try to have a look at the tests soon.

raphink:
Added tests. They fail because `aug_load_file` seems to fail silently with a fake root.

src/transform.c
```diff
@@ -822,12 +822,17 @@ int transform_load(struct augeas *aug, struct tree *xfm) {
...
     }
+
     r = filter_generate(xfm, aug->root, &nmatches, &matches);
     if (r == -1)
         return -1;
     for (int i=0; i < nmatches; i++) {
         const char *filename = matches[i] + strlen(aug->root) - 1;
         struct tree *finfo = file_info(aug, filename);
+
+        if (file != NULL && STREQ(filename, file) != 0)
```
lutter: I got the logic in my initial comment wrong; this line should read
```c
if (file != NULL && STRNEQ(filename, file))
```

That's the main reason why the test did not pass.

src/augeas.c
```diff
@@ -2020,6 +2020,43 @@ int aug_escape_name(augeas *aug, const char *in, char **out) {
...
+    struct tree *meta = tree_child_cr(aug->origin, s_augeas);
+    struct tree *load = tree_child_cr(meta, s_load);
+    char *tree_path = NULL;
+    bool found = false;
+
+    api_entry(aug);
+
+    ERR_NOMEM(load == NULL, aug);
+
+    list_for_each(xfm, load->children)  {
```
lutter: One minor thing I just noticed about this whole block: the augeas sources use 4 spaces for indentation; this block should be indented in the same style.

tests/test-api.c
```diff
@@ -663,6 +663,29 @@ static void testRm(CuTest *tc) {
...
+    aug = aug_init(root, loadpath, AUG_NO_STDINC|AUG_NO_LOAD);
+    CuAssertPtrNotNull(tc, aug);
+    CuAssertIntEquals(tc, AUG_NOERROR, aug_error(aug));
+
+    /* augeas should load a single file */
+    r = aug_load_file(aug, "/etc/fstab");
+    CuAssertRetSuccess(tc, r);
+    r = aug_get(aug, "/files/etc/fstab/1", &value);
+    CuAssertIntEquals(tc, 1, r);
+    CuAssertPtrNotNull(tc, value);
```
lutter: With that path, `value` will always be `NULL`; maybe check `/files/etc/fstab/1/spec` ?

tests/test-api.c
```diff
@@ -663,6 +663,29 @@ static void testRm(CuTest *tc) {
...
+    CuAssertRetSuccess(tc, r);
```
lutter: `CuAssertRetSuccess` checks that `r == 0`, which it will definitely not be here - just get rid of this assertion.

lutter: You should also check that we don't load more files than just `/etc/fstab`, something like this should do the trick:
```c
    r = aug_match(aug, "/files/etc/*", NULL);
    CuAssertIntEquals(tc, 1, r);
```

lutter: One final thing: we are missing the test that trying to load a nonexistent file does not produce an error; just add this to the end of `testLoadFile`:
```c
    /* augeas should return without an error when trying to load a
       nonexistant file that would be handled by a lens */
    r = aug_load_file(aug, "/etc/mtab");
    CuAssertRetSuccess(tc, r);
    r = aug_match(aug, "/files/etc/mtab", NULL);
    CuAssertIntEquals(tc, 0, r);
```

src/transform.c
raphink: Makes sense indeed. This is also why loading `/etc/fstab` resulted in `/etc/mtab` being loaded…

tests/test-api.c
raphink: Yes, I had tried with `vfstype` originally, but it didn't work (because of the bug in `transform.c`) and I forgot to put it back.

raphink:
@lutter fixed everything and squashed. Travis is now happy with it :smile:

src/augeas.c
lutter: The indentation here still needs to be adjusted to 4 spaces

src/transform.c
```diff
@@ -822,12 +822,17 @@ int transform_load(struct augeas *aug, struct tree *xfm) {
...
     }
+
     r = filter_generate(xfm, aug->root, &nmatches, &matches);
     if (r == -1)
         return -1;
     for (int i=0; i < nmatches; i++) {
         const char *filename = matches[i] + strlen(aug->root) - 1;
         struct tree *finfo = file_info(aug, filename);
+
+        if (file != NULL && STRNEQ(filename, file) != 0)
```
lutter: The condition should really just be `(file != NULL && STRNEQ(filename, file)`, i.e. without the `!= 0` - the `STRXXX` macros from `internal.h` really return boolean values (which, thanks to C, is pretty impossible to notice)

lutter:
Apart from the two minor things I commented on (indenting all of `aug_load_file` with 4 spaces and how `STRNEQ` is used in `transform_load`, this looks great. Feel free to merge after addressing those two nits. Thanks for doing this !

raphink:
@lutter done, merged, and NEWS updated. Thank you for patiently reviewing this feature.

thedrow:
Any chance to release this? This will help me a lot.

raphink:
Ideally, we had planned on releasing every 3 months, but it isn't actually really happening…

thedrow:
It'd be really useful for the osquery table. If you guys can release this. Please do.

lutter:
I plan on making an Augeas release the first week of August (Aug 5th) to finally get into the rhythm of releases every 3 months. I might be able to do it earlier, but that's doubtful.

thedrow:
@lutter Thanks for the heads up.

PR #136, opened 2014-06-16
Dnsmasq: add structure to 'address' and 'server' options

kunkku:
This is a backwards incompatible change.


raphink:
Pushed as 3e20740056bd548f714cd50e4adc6b76a249ddfc

PR #188, opened 2014-12-08
Shellvars: allow the [ builtin

kunkku:


raphink:
This is a nice and simple patch.

However, I'm wondering if wouldn't be cleaner to have a new kind of lens for this syntax. Something along the lines of 

```augeas
Shellvars.lns get "[ -f $FILENAME ]\n[[ -f $FILENAME ]]\n" =
  { @condition { "type" = "[" } { "statement" = "-f $FILENAME" } }
  { @condition { "type" = "[[" } { "statement" = "-f $FILENAME" } }
```

kunkku:
I'm afraid that is not possible with the current lens syntax. It would have to be something like:

<pre>
{ "@condition" = "-f $FILENAME" }
{ "@condition_double" = "-f $FILENAME" }
</pre>

Or alternatively:

<pre>
{ "@condition"
  { "start" = "[" }
  { "statement" = "-f $FILENAME" }
  { "end" = "]" }
}
{ "@condition"
  { "start" = "[[" }
  { "statement" = "-f $FILENAME" }
  { "end" = "]]" }
}
</pre>

Which one do you prefer? I don't especially like either...

raphink:
Or 

```augeas
{ "@condition"
  { "type" = "single" }
  { "statement" = "-f $FILENAME" }
}
{ "@condition"
  { "type" = "double" }
  { "statement" = "-f $FILENAME" }
}
```

or even

```augeas
{ "@condition"
  { "statement" = "-f $FILENAME" }
}
{ "@condition"
  { "double" }
  { "statement" = "-f $FILENAME" }
}
```
to simplify the tree (not a big fan though).

kunkku:
These suggestions suffer from the same problem as the first ones: attempting to write such lenses results in the following exception: overlapping lenses in tree union.put.

raphink:
So I went ahead and implement a version of the suggested tree from https://github.com/hercules-team/augeas/pull/188#issuecomment-72485749. The commit is 57582d491745dea4eef2c307632f89629ad352e5 and the tree is:

```augeas
  test Shellvars.lns get "[ -f $FILENAME ]\n" =
    { "@condition" = "-f $FILENAME"
      { "type" = "[" } }

  test Shellvars.lns get "[[ -f $FILENAME ]]\n" =
    { "@condition" = "-f $FILENAME"
      { "type" = "[[" } }
```

PR #192, opened 2015-01-06
Allow '+' in groupnames

grueni:
Samba users may use '+' as separator between group- and username e.g. from Active Directory


lutter:
Squashed into one commit and merged as c05e1a16

PR #193, opened 2015-01-14
PageKite: Allow more fine-grained control of 'service_on' entries

fonfon:
The definition of a pagekite 'service_on' entry is now more detailed.

Instead of the current parameters 'source' and 'destination', the new lens now allows these parameters (in addition to 'secret'): protocol, kitename, backend_host, backend_port

### DIFF
--- a/lenses/pagekite.aug
+++ b/lenses/pagekite.aug
@@ -1,5 +1,5 @@
 (*
-Module: Hostname
+Module: Pagekite
   Parses /etc/pagekite.d/
 
 Author: Michael Pimmer <blubb@fonfon.at>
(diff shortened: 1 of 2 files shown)


lutter:
The pull request looks perfectly reasonable - the only issue is that it changes the tree for Pagekite files in backwards incompatible ways. Before committing this, I'd like to give people at least a chance to object to that change. Could you start a thread on the mailing list mentioning the incompatible change and ask if anybody objects ?

fonfon:
Thanks for the hint, I just posted on the mailing list.
As the lens is pretty new I don't think that there will be any objections.

fonfon:
So far there was no reaction/objection on the mailing list

lutter:
Merged in commit ebaccad2

PR #197, opened 2015-01-20
Fixes #194 - Allow multiple interfaces in redis bind command

### LINKED ISSUE #194: Redis lense doesn't support multiple IPs in bind command

yannh:
I added the tests from mfournier, and tried to fix the tests following the help from lutter... It is important to keep in mind that you would now set bind interfaces with /files/etc/redis.conf/bind/1 instead of /files/etc/redis.conf/bind which some people might be using happily with just one interface, so this might break for some people. 

Also, I admit I'm not too sure of the difference between /files/etc/redis.conf/bind/1 and /files/etc/redis.conf/bind[1] - I failed at getting the latter version to work.

### DIFF
--- a/lenses/redis.aug
+++ b/lenses/redis.aug
@@ -54,7 +54,7 @@ redis-server.
 *)
 let standard_entry =
      let reserved_k = "save" | "rename-command" | "slaveof"
-                    | "client-output-buffer-limit"
+                    | "bind" | "client-output-buffer-limit"
   in let entry_noempty = [ indent . key k . del_ws_spc
                          . Quote.do_quote_opt_nil (store v) . eol ]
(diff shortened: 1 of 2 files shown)


mfournier:
:sparkles: :+1:

raphink:
I think I would rather use multiple `bind` nodes (resulting in  `bind[1]` instead of `bind/1`) in order to avoid breaking compatibility.

However, that might be harder to implement, as it will probably introduce an ambiguity if two `bind` nodes follow each other.

raphink:
@yannh ping?

yannh:
@raphink pong, sorry, long holidays and job switch :) Can I help?

raphink:
@yannh did you see my comments?

lutter:
Closing this in favor of PR #580.

PR #198, opened 2015-01-21
Escape special characters in path expressions

lutter:
This patch series does two things:

1. Escape characters that have special meaning in path expressions in paths that are generated/returned from Augeas (e.g. aug_match) so that they are safe to use in further Augeas calls

2. Provide an API call aug_escape_name that makes it possible to escape a string for use as one component of a path in other Augeas calls. All the characters that path expressions treat special are escaped by prefixing them with a '\\'.


lutter:
I realize that these patches have no tests - if somebody wants to beat me to it, please do; otherwise I'll try and write some tests before committing the patches.

lutter:
The revised patch series contains some tests

thoger:
This PR no longer links any commits, so adding commit ids for posterity: 1c28d67, e4f0a67, 04187f6, 7320934 and 1aa51f9.

PR #199, opened 2015-01-28
exports: add bracket support to the supported machine name formats.

vdesjardins:
Hi,

This patch add support for character class lists within [square brackets].

Thanks!


raphink:
Thank you.

PR #207, opened 2015-02-24
augtool: on interrupt, cancel current line instead of exiting

jjlin:
This patch provides more familiar shell-like handling of Ctrl-C, where the current line is cancelled instead of exiting the shell entirely. It doesn't handle modes like incremental search optimally, though. Ideally, Ctrl-C would just exit incremental search as well (behaving exactly like Ctrl-G), but in practice, it just reprints the i-search prompt. Still, you can just Ctrl-G out of it, so it's not that big a deal. Maybe someone else knows how to fix this.

### DIFF
--- a/src/augtool.c
+++ b/src/augtool.c
@@ -32,6 +32,7 @@
 #include <limits.h>
 #include <ctype.h>
 #include <locale.h>
+#include <signal.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <pwd.h>
@@ -426,6 +427,25 @@ static void print_aug_error(void) {
     }
 }
 
+static void sigint_handler(ATTRIBUTE_UNUSED int signum) {
+    // Cancel the current line of input, along with undo info for that line.
+    rl_replace_line("", 1);
+
+    // Move the cursor to the next screen line, then force a re-display.
+    rl_crlf();
+    rl_forced_update_display();
+}
+
+static void install_signal_handlers(void) {
+    // On Ctrl-C, cancel the current line (rather than exit the program).
+    struct sigaction sigint_action;
+    MEMZERO(&sigint_action, 1);
+    sigint_action.sa_handler = sigint_handler;
+    sigemptyset(&sigint_action.sa_mask);
+    sigint_action.sa_flags = 0;
+    sigaction(SIGINT, &sigint_action, NULL);
+}
+
 static int main_loop(void) {
     char *line = NULL;
     int ret = 0;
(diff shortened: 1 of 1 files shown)


raphink:
That looks very useful. @lutter, do you want to review this?

src/augtool.c
```diff
@@ -25,17 +25,18 @@
...
 #include <locale.h>
-#include <sys/types.h>
-#include <sys/stat.h>
 #include <pwd.h>
+#include <readline/history.h>
+#include <readline/readline.h>
+#include <signal.h>
 #include <stdarg.h>
+#include <sys/stat.h>
+#include <sys/types.h>
```
lutter: Why did all the includes get reordered ? That makes it hard to spot which includes were added. Could you try and restore the old order ?

src/augtool.c
```diff
@@ -426,6 +427,28 @@ static void print_aug_error(void) {
     }
 }
 
+static void sigint_handler(int unused)
+{
+    (void)unused;
```
lutter: Instead of keeping the compiler from complaining about the unused param this way, please annotate it with `ATTRIBUTE_UNUSED`, i.e. change this to

    static void sigint_handler(ATTRIBUTE_UNUSED int unused) {
      ...
    }

src/augtool.c
```diff
@@ -426,6 +427,28 @@ static void print_aug_error(void) {
...
+
+    // Move the cursor to the next screen line, then force a re-display.
+    rl_crlf();
+    rl_forced_update_display();
+}
+
+static void install_signal_handlers(void)
+{
+    // On Ctrl-C, cancel the current line (rather than exit the program).
+    struct sigaction sigint_action;
```
lutter: You should also zero out sigint_action by doing

    MEMZERO(&sigint_action, 1);

src/augtool.c
```diff
@@ -426,6 +427,28 @@ static void print_aug_error(void) {
...
+    // Cancel the current line of input, along with undo info for that line.
+    rl_replace_line("", 1);
+
+    // Move the cursor to the next screen line, then force a re-display.
+    rl_crlf();
+    rl_forced_update_display();
+}
+
+static void install_signal_handlers(void)
+{
```
lutter: It's a really minor nit, but the Augeas code base has opening braces on the same line as the function prototype

lutter:
Thanks for the patch - I agree that it is very useful, and should be committed.
I added some minor comments. Could you adress them and update the patch ? Should be good to be committed after that.

jjlin:
Okay, no problem, I can address these issues later today. The include reordering was just some incidental cleanup since the original list appeared to be in no discernible order, but I can leave that out of this patch.

jjlin:
Updated the patch to address the issues you mentioned. Please let me know if I missed anything, thanks.

lutter:
Perfect ! Just merged this as ef8e1c68. Thanks for your contribution.

PR #208, opened 2015-02-25
Added support for conf-available apache config file location

annunaki2k2:
As per the request in launchpad: https://bugs.launchpad.net/ubuntu/+source/augeas/+bug/1425183


raphink:
Merged, thank you.

PR #210, opened 2015-02-27
Added lense for AFS CellAlias

jcpunk:
This lense reads in the OpenAFS CellAlias config file and allows for simple configuration.


lutter:
Thanks for the new lens !

Merged as 35085b6f

PR #211, opened 2015-03-02
fix spelling error with chrony lense

jcpunk:
helps when I spell things correctly, updated unit test to catch in the future.


lutter:
Thanks for the fix ! Committed as 92f87380

jcpunk:
The fix is in, closing the PR

PR #212, opened 2015-03-02
Sysconfig: accept leading whitespace before variables

lutter:
Make sure we allow '  var=value' and similar entries in files

Fixes part of https://bugzilla.redhat.com/show_bug.cgi?id=761246


domcleal:
Merged as 3bc029d, thanks @lutter.

Issue #202, opened 2015-02-12
Shellvars.lns can't insert comment if the first line a newline


telnoratti:
I've tested this on CentOS 6 and CentOS 7, but I pulled the most recent version of Shellvars.lns and the issue still persists. I also haven't tested with anything other than comments, so it could be more general.
### Steps to reproduce
Create a file that is valid shell syntax with a newline at the beginning called test. https://linx.li/b59d7e0.txt is what I used.
```
augtool -A -L
set /augeas/load/Shellvars/lens "Shellvars.lns"
set /augeas/load/Shellvars/incl "/test"
load
ins #comment before /files/test/*[1]
set /files/test/#comment[1] "test comment"
save
```
I get the following error when printed
```
/augeas/files/test/error = "put_failed"
/augeas/files/test/error/path = "/files/test"
/augeas/files/test/error/lens = "/usr/share/augeas/lenses/dist/shellvars.aug:191.12-.56:"
[...]

PR #214, opened 2015-03-05
Shellvars: fix insertion at top of file that starts with blank lines

lutter:
Possible blank lines at the start of a file were mapped with Util.empty
which produces empty 'hidden' nodes (nodes with NULL label and value)

This causes problem when a user inserts something before these nodes, for
example with "insert 'foo' before /*[1]". Even though the nodes that
Util.empty produces are generally hidden from the user, they are still in
the tree and /*[1] selects such a node. The resulting tree has a hidden
node in the middle, something that Shellvars.lns does not allow and
therefore putting this tree fails. This is really hard to understand for
users as the tree they see with 'print' is indistinguishable from one that
doesn't have these hidden nodes.

To fix this, the changed lens simply deletes the blank lines at the
beginning of the file without creating any tree nodes. This makes the above
insertion work at the cost of losing these blank lines as soon as anything
is inserted into or deleted from the file.

Fixes https://github.com/hercules-team/augeas/issues/202

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,10 @@
   - Lens changes/additions
     * Dnsmasq: Parse the structure of the 'address' and 'server' options
                (incompatible change) (Kaarle Ritvanen)
+    * Shellvars: make insertion at the beginning of a file that starts with
+                 blank lines work; the new lens will remove blank lines
+                 from the beginning of a file as soon as lines are added to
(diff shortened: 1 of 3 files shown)


domcleal:
Seems a fair enough solution.  Merged as be4a905, thanks @lutter.

PR #216, opened 2015-03-11
Added iscsid lense per 174

jcpunk:
While I don't have much use for this lense, it was easy enough to add in the unit tests.


lutter:
Thanks for the new lens.

Merged as 6239943f

Issue #179, opened 2014-11-25
nginx lens doesn't support recursive blocks


jeffNagra:
As discussed with Raphink....

The nginx lens is failing the parse even the default nginx.conf supplied with the nginx package (below)…
 
set /augeas/load/nginx/lens "Nginx.lns"
set /augeas/load/nginx/incl "/home/root/jeff"
load
print /augeas//error

/augeas/files/home/root/jeff/error = "parse_failed"
/augeas/files/home/root/jeff/error/pos = "410"
/augeas/files/home/root/jeff/error/line = "20"
/augeas/files/home/root/jeff/error/char = "0"
/augeas/files/home/root/jeff/error/lens = "/usr/share/augeas/lenses/dist/nginx.aug:61.10-.57:"
/augeas/files/home/root/jeff/error/message = "Iterated lens matched less than it should"

Full default nginx.conf (comes with the nginx RPM):

```
# For more information on configuration, see:
#   * Official English Documentation: http://nginx.org/en/docs/
#   * Official Russian Documentation: http://nginx.org/ru/docs/

user  nginx;
worker_processes  1;

error_log  /var/log/nginx/error.log;
#error_log  /var/log/nginx/error.log  notice;
#error_log  /var/log/nginx/error.log  info;

pid        /run/nginx.pid;


events {
    worker_connections  1024;
}


http {
    include       /etc/nginx/mime.types;
    default_type  application/octet-stream;

    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
                      '$status $body_bytes_sent "$http_referer" '
                      '"$http_user_agent" "$http_x_forwarded_for"';

    access_log  /var/log/nginx/access.log  main;

    sendfile        on;
    #tcp_nopush     on;

    #keepalive_timeout  0;
    keepalive_timeout  65;

    #gzip  on;

    index   index.html index.htm;

    # Load modular configuration files from the /etc/nginx/conf.d directory.
    # See http://nginx.org/en/docs/ngx_core_module.html#include
    # for more information.
    include /etc/nginx/conf.d/*.conf;

    server {
        listen       80 default_server;
        server_name  localhost;
        root         /usr/share/nginx/html;

        #charset koi8-r;

        #access_log  /var/log/nginx/host.access.log  main;

        # Load configuration files for the default server block.
        include /etc/nginx/default.d/*.conf;

        location / {
        }

        # redirect server error pages to the static page /40x.html
        #
        error_page  404              /404.html;
[...]

PR #225, opened 2015-04-09
Nginx: rework to support a much broader set of valid configurations

lutter:
This patch should fix Issue #179

### DIFF
--- a/lenses/nginx.aug
+++ b/lenses/nginx.aug
@@ -33,28 +33,76 @@ autoload xfm
 
 (* Variable: block_re
      The keywords reserved for block entries *)
-let block_re = "http" | "events"
-             | "server" | "upstream"
-             | "location" | "if"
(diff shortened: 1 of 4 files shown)


lenses/nginx.aug
```diff
@@ -33,28 +33,43 @@ autoload xfm
 
 (* Variable: block_re
      The keywords reserved for block entries *)
-let block_re = "http" | "events"
-             | "server" | "upstream"
-             | "location" | "if"
+let block_re = "http" | "events" | "server" | "upstream"
+
```
diracdeltas: I think this is missing some block types, like `mail`

lutter: Thanks for pointing this out.

I've gone through the nginx docs now and added the following block types: mail, stream, geo, map, and split_clients

I *hope* that's all of them. If not, please let me know.

PR #226, opened 2015-04-10
Much better error message on 'short iteration'

lutter:
The error message produced when an iterated lens `l*` does not match the input text was not very helpful, since it didn't give any indication where in `l` we got stuck. With this patch, Augeas tries to pinpoint where exactly `l` does not match by going through `l`'s sublenses and finding the left-most lens that does match and the next one after that that doesn't match.

A typical error message before this change was

    hocon_lite.aug:29.5-31.0:exception: Iterated lens matched less than it should
        Lens: lenses/hocon_lite.aug:24.10-.26:
        Error encountered at 1:0 (0 characters into string)
                            <|=|master: {\n    allow: false\n}\n>

With this change the error message becomes

    hocon_lite.aug:29.5-31.0:exception: Iterated lens matched less than it should
        Lens: lenses/hocon_lite.aug:24.10-.26:
          Last match: /homes/lutter/code/augeas/lenses/build.aug:285.3-.61:
          Not matching: /homes/lutter/code/augeas/lenses/build.aug:286.3-.57:
        Error encountered at 2:0 (10 characters into string)
                        <master: {\n|=|    allow: false\n}\n>

Having the last matching and the next non-matching lens, plus the more precise position `|=|` made it really easy to figure out in this case what was wrong with the lens.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -2045,7 +2045,7 @@ void aug_close(struct augeas *aug) {
 
 int __aug_load_module_file(struct augeas *aug, const char *filename) {
     api_entry(aug);
-    int r = load_module_file(aug, filename);
+    int r = load_module_file(aug, filename, NULL);
     api_exit(aug);
     return r;
 }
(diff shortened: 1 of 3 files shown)


lutter:
After thinking some more about the errors we produce, I added a 'errors' command to augrun and tried to improve the way atypes get printed, too

PR #228, opened 2015-04-14
Improvements to Jaas lens

sshipway:
Add flexibility to Jaas lens.  Allow comments within Modules, optionless flags, options without intervening linebreaks, and true/false option values without quotes.  All of these are valid syntax in the Shibboleth file, and prevent our live file from parsing correctly.

The existing lens still has problems with comments within comments, but this is fairly rare.


sshipway:
Works against our tests, but not against the official tests.  Partly because the new lens does not include the ';' statement terminator in the option text (which I think was an error in the original) but also due to a separate parse error.  Investigating it now, will resubmit request after it is rectified.

PR #230, opened 2015-04-14
Fixes and Improvements to Jaas lens

sshipway:
Tested against augparse successfully.  Requires latest Util.aug to function correctly.

This corrects an error in the existing jaas.aug lens (terminating ; was included in option value), plus adds support for several other syntactically valid formatting options (comments within option lists, options on the same line, uncommented boolean values, flags without options)

An updated test script tests for the newly supported syntax combinations, and corrects for the semicolon error.

This replaces my previous pull request which did not take into account other changes in the Util.pm and the test script.


lutter:
Thanks for the update ! Merged as d17f2fbc

PR #231, opened 2015-04-15
Add rpm default paths to postgresql.aug

liuhaotian:
As of Postgres 9.4, postgresql.conf is stored in the database cluster's
data directory

See:
http://www.postgresql.org/docs/9.4/static/runtime-config-file-locations.html


lutter:
Thanks for the patch ! Committed as e733a54f

PR #233, opened 2015-04-16
Limits lens: Allowing comments at end of line

timdeluxe:
Here is my proposal for changing limits.aug lens to allow comments at the end of the line, which are no problem for the OS and heavily used in the limits.conf on our systems. Without that change you get a parse error.


lutter:
Thanks for the patch !
I added a test and committed it as 96c47552

PR #241, opened 2015-05-16
Add passwd lens support for nis [+-]username syntax

stoicbo:
The /etc/passwd file can contain nis users with the syntax [+-]username for the pruprose of including or excluding specific accounts.

Importing of particular users' accounts can be restricted by using the prefix "+" with a username.

root:x:0:0:root user:/home/root:/bin/bash
+bob::::::
+alice::::Alice:/home/alice:/bin/bash

Certain users' accounts can be excluded from being imported by using the "-" prefix with a username.

root:x:0:0:root user:/home/root:/bin/bash
-bob::::::
+

The string that can copmrise a valid username has been updated to cover a range slightly larger that the definitions in POSIX (http://pubs.opengroup.org/onlinepubs/9699919799/ , section 3.431), is_valid_name() in chkname.c in shadow-utils and the NAME_REGEX definition in distributions where defined. Rx.word was too broad and did not allow for the -username syntax.

### DIFF
--- a/lenses/passwd.aug
+++ b/lenses/passwd.aug
@@ -28,7 +28,7 @@ let comment    = Util.comment
 let empty      = Util.empty
 let dels       = Util.del_str
 
-let word       = Rx.word
+let word       = /[_.A-Za-z0-9][-\@_.A-Za-z0-9]*\$?/
 let integer    = Rx.integer
 
 let colon      = Sep.colon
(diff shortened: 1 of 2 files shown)


lenses/passwd.aug
```diff
@@ -28,7 +28,7 @@ let comment    = Util.comment
 let empty      = Util.empty
 let dels       = Util.del_str
 
-let word       = Rx.word
+let word       = /[_.A-Za-z0-9][-\@_.A-Za-z0-9]*\$?/
```
lutter: Why is `$` allowed at the end of this regexp ? From further down, it seems to be pulling double-duty as a regexp for passwords, and for user names.

stoicbo:
In shadow-utils, is_valid_name() in chkname.c defines the following:
/*
* User/group names must match [a-z_][a-z0-9_-]*[$]
*/
What I was able to find out is that the $ at the end is allowed for samba conformity.

For the password field I used the definition in nisdefault. This is probably wrong. I will refactor for the common sections in nisentry, nisuserplus, nisuserminus and address username/password

stoicbo:
Updated to address the username and password fields. Updated the nis section to consolidate common sections.

lutter:
Reformatted commit message, squashed both commits into one, added to NEWS file and committed as a5eea785.

Thanks for doing this !

PR #243, opened 2015-05-21
Updated tests/tes-put-symlink-augsave.sh to remedy failure on Solaris

ggardella:
The output from Solaris 'sum' command was causing this test to fail as it included the file path in the output.


ggardella:
Test passes on Solaris now.

domcleal:
Merged as dc62b24, thanks @ggardella!

PR #244, opened 2015-05-22
RFC: Add a 'region' lens to make controlling spacing easier

lutter:
This is not meant to be committed yet, but I would really appreciate feedback based on people's experimenting with these patches.

So far, how deleted text is handled when entries are added to or removed
from the tree is tightly bound up with how the tree was structured,
because the subtree combinator [ .. ] controlled both: construction of new
tree nodes and glomming deleted stuff together.

In addition, the [ .. ] does the glomming based on the key of the tree node
it constructs, so that deleting a node from hte middle of a number of nodes
with the same key had the weird effect that entries tended to 'shift up'.

The region combinator tries to address this, in that all it does is
demarcate how deleted stuff should be glommed, and it does that based on
the position in the input file, not on the value of a key.

Syntactically, this is written as < l >, though other syntax, like 'region
l', or '<< l >>' might be better.

It would be great if this could also be used to indent new things the same
as surrounding stuff, i.e. if we have a lens < l >* and we add en entry in
the middle of it, if that would automatically mean that the new entry gets
the indentation of one of the surrounding lines - that's not implemented
yet.

I would really like to see and hear how this does or does not address some
of the spacing/indentation problems that people have been having, ideally
demonstrated by a simple test.

### DIFF
--- a/src/ast.c
+++ b/src/ast.c
@@ -93,6 +93,21 @@ struct dict *make_dict(char *key, struct skel *skel, struct dict *subdict) {
     return NULL;
 }
 
+static char *size_as_key(size_t key) {
+    char *s = NULL;
+
+    xasprintf(&s, "%zx", key);
+    return s;
+}
+
+struct dict *make_dictz(size_t key, struct skel *skel, struct dict *subdict) {
+    char *s = size_as_key(key);
+
+    if (s == NULL)
+        return NULL;
+    return make_dict(s, skel, subdict);
+}
+
 void free_dict(struct dict *dict) {
     if (dict == NULL)
         return;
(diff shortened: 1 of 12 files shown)


raphink:
I think the feature is really interesting, and would solve a lot of issues (although I'd be more in favor of eventually providing an API exposing the deleted elements so we can control them).

I haven't given it too much thought, but I can't think of a case where it wouldn't be good to make the region behavior the default, instead of patching all the existing lenses.

lutter:
I just rebased this to latest master.

I am a little hesitant to just change the behavior of `[ .. ]` wholesale, though it's certainly an option; I wonder though that in lenses where `[..]` is used on parts of a line we'd get funny spacing behavior, for example in `hosts.aug` when you delete an alias.

raphink:
Do all the tests pass if you change the default behavior?

Also, does it make sense to keep `seq` entries with region lenses, or could we just get back to static labels, which makes tree manipulation much easier?

lutter:
Updated to make it possible to switch between the current behavior of `[...]` and what `<...>` in this PR did by setting the environment variable `AUGEAS_NO_SHIFT`. When this variable is set, lenses should not exhibit the 'shifting up' behavior that they usually have any more, but we will lose idempotency to some extent. This update is mostly to make it possible to try this new behavior out inrelation to Issue #450.

It's expected that the tests for this PR fail as-is as I added a test that only works when `AUGEAS_NO_SHIFT` is set.

Issue #113, opened 2014-04-11
will not parse my.cnf if any !include directives are present


ss8913:
In 0.9.0, my my.cnf parses fine.  With 1.0.0 or newer, it does not.  If I take out the !include /sub/path.d/ directive, it works, but obviously I need that in there.  I saw in the changelog that something changed regarding !include directives, but it appears to have broken this functionality and is locking me into 0.9.0.

lutter:
Is this still an issue ?

ss8913:
Yes, I actually tested it yesterday with the latest augeas on one of my
Centos 6.6 hosts.  Still returns nothing when you do:

augtool print /files/etc/my.cnf

As long as the !include is present in the file.  Without !include, it works
fine.  Versions of augeas before 1.0.0 work fine either way.

lutter:
The issue is actually not the include directive itself, but that Augeas parses that in a way that ends whatever section it appears in. The error happens on the line starting with `port` right after the `!includedir` line.

Not sure how to fix that though .. we'd probably have to accept that line as part of the `mysqld` section

ss8913:
However it works in 0.9.0, it was OK then :)  Let me know if you want me to
test this for you, I have thousands of boxes that I can reproduce it on,
and I can send you one of the my.cnf files as well if I haven't already.

PR #245, opened 2015-05-22
MySQL (incompatible change): handle !include directives in the middle of sections

lutter:
MySQL (incompatible change): handle !include directives in the middle…

CentOS 6.6 (at least) ships with a mysql.conf that has a !include directive
in the middle of a section. To accomodate that, we pull !include directives
into the sections in which they appear rather than having htem end the
section. That's the incompatible change: these directives now appear
underneath /target/!include rather than at /!include

Fixes https://github.com/hercules-team/augeas/issues/113

### DIFF
--- a/lenses/mysql.aug
+++ b/lenses/mysql.aug
@@ -15,12 +15,15 @@ let comment  = IniFile.comment IniFile.comment_re "#"
 let sep      = IniFile.sep IniFile.sep_re IniFile.sep_default
 
 let entry    =
-     let bare = Quote.do_dquote_opt_nil (store /[^#;" \t\r\n]+([ \t]+[^#;" \t\r\n]+)*/)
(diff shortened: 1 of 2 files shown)


raphink:
Is that really necessary though?

lutter:
I can't come up with another way to allow this sort of format. Would love to see a backwards compatible fix. Do you have any ideas on how ?

raphink:
I'll have a look

raphink:
I don't really get the issue. If I add `!include` statements in the middle of the file, they get parsed OK:

https://gist.github.com/raphink/f075edcb59b1922cf899

raphink:
Ah sorry @lutter, @domcleal just explained to me what you meant… I see the issue now…

I guess there would be an option of allowing include statements inside sections but not at the end of the section, but that wouldn't be very intuitive. So if that's not an option, your fix is the only way.

raphink:
@lutter 	should we merge this then?

PR #246, opened 2015-05-26
integrate two patches from FreeBSD ports

mmoll:
- /boot/loader.conf can be parsed with the sysctl lens
- /etc/perodic.conf can be parsed with the shellvars lens

Variations of these patches are floating around in the FreeBSD ports tree for quite some time, lastly updated in freebsd/freebsd-ports@a378156b, but I'd really like to get these upstreamed.


domcleal:
Thanks @mmoll, merged as 754ad5d!

PR #248, opened 2015-05-29
Chrony: update for chrony 2.0

mlichvar:
In chrony-2.0 was added a new directive to specify a pool of NTP servers
and there are also several new options that can be used with the
server/peer/pool directives.

### DIFF
--- a/lenses/chrony.aug
+++ b/lenses/chrony.aug
@@ -9,13 +9,6 @@ About: Reference
 
   See http://chrony.tuxfamily.org/manual.html#Configuration-file
 
-About: Limitations
-  Does not (currently) support
-    - include
-    - manual
-    - refclock
-    - tempcomp
-
 About: License
   This file is licenced under the LGPL v2+, like the rest of Augeas.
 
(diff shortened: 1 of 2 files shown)


domcleal:
There are a few other options which I noted over at https://bugzilla.redhat.com/show_bug.cgi?id=1213281#c3 from the [NEWS file](http://git.tuxfamily.org/chrony/chrony.git/tree/NEWS), such as minsources, smoothtime and tempcomp.  I wonder if you'd mind adding those too?

mlichvar:
I can look into that. There are other directives that were not supported before chrony 2.0, like refclock, manual, include, and which are listed as limitations in the chrony aug file. Should be all obscure directives supported?

domcleal:
Thanks.  It'd be great if they were, since it would prevent the file from being parsed at all if they're used.

mlichvar:
I've reworked the patch to add all missing directives and options as supported by current chrony git. I've also made some fixes to allow indentation, not allow comment on end of line and correctly parse broadcast with port.

domcleal:
That's excellent, thanks @mlichvar!

mlichvar:
Thanks!

PR #250, opened 2015-06-01
Fix passwd lens for @nisdefault on OpenBSD

bodgit:
Following `passwd(5)` on OpenBSD results in an `/etc/passwd` NIS entry like:

```
+:*:0:0:::
```

Passwd lens barfs on the '*' password. Change to match how normal entries are
matched.


lutter:
Reformatted commit message, made small change to remove incompatibility with empty password, added another test and committed as 1128acba

PR #252, opened 2015-06-02
Parse OpenBSD's hostname.if(5) files with the simplelines lens

jasperla:


lutter:
Thanks for the patch. Merged as 8251eb3e

Issue #256, opened 2015-06-10
On OSX libedit lacks rl_crlf and rl_replace_line


lutter:
On OSX, readline isn't available by default; instead, augtool gets linked against libedit. That is missing a few functions (rl_crlf and rl_replace_line) which need to be handled in the code similar to how rl_completion_matches is handled

dtrebbien:
One way of working around this issue is to install the [Homebrew](http://brew.sh) readline package and then run `./autogen.sh` with the following environment variables set:

<pre>
CPPFLAGS=-I/usr/local/opt/readline/include
LDFLAGS=-L/usr/local/opt/readline/lib
</pre>

(tested Mac OS X 'El Capitan' version 10.11.1)

I think that this approach is preferable to adding replacements (PR #257) because readline would actually be used instead of the BSD libedit library.

lutter:
This workaround wouldn't keep you from using the Homebrew libraries though, right ? I.e., even with this patch committed, if you do what you said above you would get a fully functional readline, right ?

The reason why I'd like to commit this is that the buildsystem where this became an issue doesn't have the Homebrew readline AFAIK.

dtrebbien:
> This workaround wouldn't keep you from using the Homebrew libraries though, right ? I.e., even with this patch committed, if you do what you said above you would get a fully functional readline, right ?

Yes, you are right.

> The reason why I'd like to commit this is that the buildsystem where this became an issue doesn't have the Homebrew readline AFAIK.

Ah, I see.

PR #257, opened 2015-06-11
* src/augtool.c: add stubs for readline functions missing on OSX

lutter:
OSX ships with a crippled version of readline, libedit, which does not
contain rl_crlf and rl_replace_line. This patch makes builds on OSX work
again by providing trivial replacements for these functions.

Fixes https://github.com/hercules-team/augeas/issues/256


haus:
:+1:

PR #258, opened 2015-06-15
Add support for /etc/master.passwd

bodgit:
Found on OpenBSD and possibly other BSD's.

Also update /etc/group lens with support for NIS map with an overridden and
disabled password, i.e. `+:*::`


raphink:
LGTM.

raphink:
Just one thing. I'd rather see the changes in `Group` in a separate commit from the one creating the new lens.

bodgit:
Fixed

raphink:
Merged as fd9cf8c..51d4879

PR #259, opened 2015-06-30
Fixes "server" entries in upstream blocks

jtopjian:
The "server" keyword was reserved for blocks and therefore excluded
from normal entries. This commit adds "server" as a valid entry.

### DIFF
--- a/lenses/nginx.aug
+++ b/lenses/nginx.aug
@@ -42,7 +42,7 @@ let block_re_all = block_re | "if" | "location" | "geo" | "map"
 (* View: simple
      A simple entry *)
 let simple =
-     let kw = Rx.word - block_re_all
+     let kw = Rx.word - block_re_all | "server"
   in let sto = store /[^ \t\n;][^;]*/ . Sep.semicolon
   in [ Util.indent . key kw . Sep.space . sto . (Util.eol|Util.comment_eol) ]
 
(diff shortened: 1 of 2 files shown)


jtopjian:
Please let me know if there's a better way of doing this. :)

raphink:
Thank you for that suggestion. Your test happened to pass because a non-recursion test was missing for that lens. Can you rebase on master please?

jtopjian:
Rebasing on master invalidates the changes I made. 

What is the next best way of allowing the `server` keyword inside an `upstream` block?

raphink:
So `server` is normally a block entry, but it can be used a simple entry in `upstream` blocks, is that right?

jtopjian:
That's correct. I'd even say that `server` is the most important keyword of the `upstream` block:

http://nginx.org/en/docs/http/ngx_http_upstream_module.html

Normally if there would be some type of workaround by using another block or keyword, I'd opt for that, but `server` is kind of important in this case :)

raphink:
We might want to look again at the actual syntax. It's currently mapped as a recursive block syntax, but I don't think that's actually correct.

jtopjian:
Ah, I see. So it sounds like the correct way to implement this would involve a lot more than a simple fix? If so, then that's definitely outside of my Augeas knowledge (at least for now) and I'd be happy to close this PR.

raphink:
Fixed in 2b14a809226a756d6cc12ba95241686d96ce2f8b

PR #261, opened 2015-07-03
Rhsm: new lens to parse subscription-manager's rhsm.conf

domcleal:


lenses/rhsm.aug
```diff
@@ -0,0 +1,30 @@
...
+  This lens applies to:
+    /etc/rhsm/rhsm.conf
+
+  See <filter>.
+*)
+
+module Rhsm =
+  autoload xfm
+
+let lns     = IniFile.lns_loose_multiline
```
raphink: As said on IRC, I'd rather use `IniFile.lns_loose*` lenses for guessing work, not specific lenses.

domcleal: Thanks for the comment, I've updated it to be precise about what Python's ConfigParser supports.

raphink:
Looking good, want to merge it?

domcleal:
Thanks for reviewing, abdb9fb.

PR #263, opened 2015-07-10
Support for negated command alias in /etc/sudoers - fixes #262

### LINKED ISSUE #262: Sudoers lens doesn't support `!`
Augeas Team,

Recently we got issues with sudoers lens (sudoers.aug) doesn't support the addition of a `!` used to negate values in a default /etc/sudoers file.
Eg: !DISALLOWED 
Throws out errors:
Debug: Augeas[includedirsudoers](provider=augeas): /augeas/files/etc/sudoers/error/message = Iterated lens matched less than it should 

As such, we can't proceed to use this lens which we need urgently. 
Please assist us with a solution/workaround asap as we suspect this could be a bug.

Thanks,
Anandk

GeoffWilliams:
Hi there,

This fixes the sudoers lens to support negating command alias groups like this:
```
 %sportshorseracingopssudoers ALL=(ALL) ALL, !DISALLOWED
```

Which currently fails if the command alias appears _after_ a comma.

An additional unit test has been added to support this fix which is also documented at https://fedorahosted.org/augeas/ticket/346 (along with a patch which was a useful confirmation that the fix I'd come up with was correct)

How does this look?  Would be great to get this fixed once and for all.

Thanks,
Geoff

### DIFF
--- a/lenses/sudoers.aug
+++ b/lenses/sudoers.aug
@@ -82,15 +82,46 @@ let sep_col  = sep_cont_opt_build ":"
 (* Variable: sep_dquote *)
 let sep_dquote   = Util.del_str "\""
 
+(************************************************************************
+ * View: default_type
+ *   Type definition for <defaults>
+ *
+ *   Definition:
+ *     > Default_Type ::= 'Defaults' |
+ *     >                  'Defaults' '@' Host_List |
+ *     >                  'Defaults' ':' User_List |
+ *     >                  'Defaults' '!' Cmnd_List |
+ *     >                  'Defaults' '>' Runas_List
+ *************************************************************************)
+let default_type     =
+  let value = store /[@:!>][^ \t\n\\]+/ in
+  [ label "type" . value ]
+
+(************************************************************************
+ * View: del_negate
+ *   Delete an even number of '!' signs
+ *************************************************************************)
+let del_negate = del /(!!)*/ ""
+
+(************************************************************************
+ * View: negate_node
+ *   Negation of boolean values for <defaults>. Accept one optional '!'
+ *   and produce a 'negate' node if there is one.
+ *************************************************************************)
+let negate_node = [ del "!" "!" . label "negate" ]
+
+let negate_or_value (key:lens) (value:lens) =
+  [ del_negate . (negate_node . key | key . value) ]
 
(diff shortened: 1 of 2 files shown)


lenses/tests/test_sudoers.aug
```diff
@@ -326,3 +326,16 @@ test Sudoers.spec get "group+user somehost = ALL\n" =
       { "command" = "ALL" }
     }
   }
+
+(* Test: Sudoers.spec
+     https://github.com/hercules-team/augeas/issues/262:  Sudoers lens doesn't suppot `!` for command aliases *)
```
raphink: typo in comment

lenses/sudoers.aug
```diff
@@ -88,7 +88,7 @@ let sep_dquote   = Util.del_str "\""
 (* Variable: sto_to_com_cmnd
 sto_to_com_cmnd does not begin or end with a space *)
 let sto_to_com_cmnd =
-      let alias = Rx.word - /(NO)?(PASSWD|EXEC|SETENV)/
+      let alias = /!?/ . Rx.word - /(NO)?(PASSWD|EXEC|SETENV)/
```
raphink: I'd rather use the `negate` and `negate_node` lenses, so it behaves like the rest of the lens:

```augeas
let alias = del_negate . negate_node? . Rx.word - /(NO)?(PASSWD|EXEC|SETENV)/
```

GeoffWilliams:
Hi Raphink,

Thanks for looking at this.  I've updated the PR as you asked.  To get the `negate` and `negate_node` lenses to work I had to re-arrange the line a little or I got an error about concatenating a lens with a regexp.

I also had to move some definitions to be earlier in the file so that I could reference them.  I've updated the tests to reflect the `{ "negate" }` node that now gets created and have also updated the definition of `non_alias` as this previously had a hardcoded regexp for a single `!` character.

raphink:
I understand the reorganization due to lenses needing other lenses. However, it means that some defaults definitions end up outside the defaults group, which will generate an ugly doc. Would it be possible to exchange the two groups altogether maybe?

raphink:
Merged as 58a009f519778cc4810dbbcbe03657038bb7e6e1

PR #266, opened 2015-07-15
Keepalived updates

jtopjian:
I didn't review any of the virtual server stuff since I'm not working with it at the moment.

As always, please let me know if there are better ways of achieving any of this. :smile:

### DIFF
--- a/lenses/keepalived.aug
+++ b/lenses/keepalived.aug
@@ -215,7 +215,7 @@ let vrrp_instance = named_block "vrrp_instance" vrrp_instance_field
 
 (* View: vrrp_script_field *)
 let vrrp_script_field =
-      let num_re = "interval" | "weight"
+      let num_re = "interval" | "weight" | "fall" | "raise"
    in let to_eol_re = "script"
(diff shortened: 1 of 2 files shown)


raphink:
This looks good (although I'm afraid it's going to make this lens —probably already the heaviest one— even heavier). Can you just squash some commits please?

jtopjian:
I've noticed that this lens is a bit slow -- is that what you mean by "heavy"?

Sure can. The multiple commits were to just make a code review easier. 

Is it OK if I just squash everything into 1 commit with a summary? That'll probably be the cleanest with all of the spacing fixes.

raphink:
I would rather you squash the "more of that" commits into the "that" original ones, e.g. "More spacing/tabs" into "Tabs and spacing fixes".

Also, I'd appreciate if you could prepend "Keepalived: " in front of each commit message.

jtopjian:
OK! Thank you for the suggestions. This PR looks much cleaner now :)

raphink:
Merged. Thank you.

PR #270, opened 2015-07-30
fstab: remove unneeded brackets

roidelapluie:


raphink:
Fair enough…

raphink:
Merged as 5628653

Issue #272, opened 2015-07-31
ssh_config is not parsed correctly with recent options


omgold:
I had two problems with the ssh (/etc/ssh_config) lense. At first it didn't accept indents on lines for Ciphers, MACs, etc., second it didn't split by comma the new options HostKeyAlgorithms and KexAlgorithms. I created a patch to fix that.

omgold:
Here is the fix:

--- /usr/share/augeas/lenses/dist/ssh.aug	2015-07-31 10:41:00.244537819 +0200
+++ ssh.aug	2015-07-31 10:43:22.666885846 +0200
@@ -51,7 +51,7 @@
 
     let commas_entry (k:regexp) =
          let value = [ seq "commas_entry" . value_to_comma]
-      in [ key k . counter "commas_entry" . spc .
+      in [ indent . key k . counter "commas_entry" . spc .
            Build.opt_list value comma . eol ]
 
     let fw_entry (k:regexp) = [ indent . key k . spc .
@@ -66,6 +66,8 @@
 
     let ciphers = commas_entry /Ciphers/i
     let macs	= commas_entry /MACs/i
+    let hostkeyalgorithms = commas_entry /HostKeyAlgorithms/i
+    let kexalgorithms	= commas_entry /KexAlgorithms/i
 
     let special_entry = send_env
 	                    | proxy_command
@@ -73,6 +75,8 @@
 	                    | local_fw
 	                    | macs
 	                    | ciphers
+	                    | hostkeyalgorithms
+	                    | kexalgorithms
 
     let key_re = /[A-Za-z0-9]+/
                - /SendEnv|Host|ProxyCommand|RemoteForward|LocalForward|MACs|Ciphers/i

domcleal:
Could you rebase your patch against master and open a pull request please?  It looks to be against an old version, as master currently does parse KexAlgorithms as a list, but not HostKeyAlgorithms.

omgold:
Can't see KexAlgorithms in master, but I will do the rebase.

domcleal:
https://github.com/hercules-team/augeas/blob/master/lenses/sshd.aug#L115

domcleal:
Ah, Ssh versus Sshd modules, apologies.

raphink:
This patch would be more efficient as one entry:

```augeas
let algorithms = commas_entry /(Kex|HostKey)Algorithms/i
```

raphink:
[...]

PR #273, opened 2015-07-31
fixing ssh lense, as mentioned in issue #272

omgold:
- accept indents on lines for Ciphers, MACs, etc.
- split by command the options HostKeyAlgorithms and KexAlgorithms

### DIFF
--- a/lenses/ssh.aug
+++ b/lenses/ssh.aug
@@ -66,6 +66,8 @@ module Ssh =
 
     let ciphers = commas_entry /Ciphers/i
     let macs	= commas_entry /MACs/i
+    let hostkeyalgorithms = commas_entry /HostKeyAlgorithms/i
+    let kexalgorithms = commas_entry /KexAlgorithms/i
 
     let special_entry = send_env
 	                    | proxy_command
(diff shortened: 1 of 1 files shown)


lenses/ssh.aug
```diff
@@ -66,6 +66,8 @@ module Ssh =
 
     let ciphers = commas_entry /Ciphers/i
     let macs	= commas_entry /MACs/i
+    let hostkeyalgorithms = commas_entry /HostKeyAlgorithms/i
+    let kexalgorithms = commas_entry /KexAlgorithms/i
```
raphink: For performance reasons, I would rather make one entry for those:

```augeas
let algorithms = commas_entry /(HostKey)|(Kex)Algorithms/i
```

lenses/ssh.aug
```diff
@@ -73,9 +75,11 @@ module Ssh =
 	                    | local_fw
 	                    | macs
 	                    | ciphers
+	                    | hostkeyalgorithms
+	                    | kexalgorithms
 
     let key_re = /[A-Za-z0-9]+/
-               - /SendEnv|Host|ProxyCommand|RemoteForward|LocalForward|MACs|Ciphers/i
+               - /SendEnv|Host|ProxyCommand|RemoteForward|LocalForward|MACs|Ciphers|HostKeyAlgorithms|KexAlgorithms/i
```
raphink: And then here again, using the same regexp as above.

raphink:
Could you add unit tests for these entries please?

raphink:
You could have just pushed forced… especially since you haven't made a feature branch anyway…

Issue #271, opened 2015-07-30
Support for termcap-style capability databases


bodgit:
It would be handy to support termcap-style capability database files. This generic format is used in a number of configuration files; `/etc/printcap`, `/etc/login.conf`, `/usr/share/misc/termcap` to name but a few.

The format consists blank lines, `#` comments and records which are `:`-separated, the first field is the name followed by one or more capabilities. What makes it trickier is that a record can be split across multiple lines with `\` to mark continuation and extra `:`'s can be added for clarity, any capability that consists solely of whitespace are ignored.

For example the following two examples are considered equivalent:

```
# A comment

authpf:\
        :welcome=/etc/motd.authpf:\
        :shell=/usr/sbin/authpf:\
        :tc=default:
```

```
# A comment

authpf:welcome=/etc/motd.authpf:shell=/usr/sbin/authpf:tc=default:
```

Other examples from various files are:

```
default:\
        :path=/usr/bin /bin /usr/sbin /sbin /usr/X11R6/bin /usr/local/bin /usr/local/sbin:\
        :umask=022:\
        :datasize-max=512M:\
        :datasize-cur=512M:\
        :maxproc-max=256:\
        :maxproc-cur=128:\
        :openfiles-cur=512:\
        :stacksize-cur=4M:\
        :localcipher=blowfish,8:\
        :ypcipher=old:\
        :tc=auth-defaults:\
        :tc=auth-ftp-defaults:
```

```
example|an example of binding multiple values to names:\
     :foo%bar:foo^blah:foo@:\
     :abc%xyz:abc^frap:abc$@:\
     :tc=more:
```

```
dumb|80-column dumb tty:\
        :am:\
        :co#80:\
        :bl=^G:cr=^M:do=^J:sf=^J:
unknown|unknown terminal type:\
        :gn:tc=dumb:
lpr|printer|line printer:\
        :bs:hc:os:\
        :co#132:li#66:\
        :bl=^G:cr=^M:do=^J:ff=^L:le=^H:sf=^J:
glasstty|classic glass tty interpreting ASCII control characters:\
        :am:bs:\
        :co#80:\
        :bl=^G:cl=^L:cr=^M:do=^J:kd=^J:kl=^H:le=^H:nw=^M^J:ta=^I:
```

I would expect augeas to parse one of these records to something like:

```
{ "example|an example of binding multiple values to names"
  { "foo%bar" }
  { "foo^blah" }
  { "foo@" }
  { "abc%xyz" }
[...]

PR #274, opened 2015-07-31
Add lens for reading termcap-style databases

bodgit:
This is my stab at fixing #271 which seems to parse everything I've thrown at it including a ~900 KB termcap file. I can also add new records to files. However unless I disable type checking I get:

```
./getcap.aug:34.2-.62:Failed to compile record
./getcap.aug:34.43-.54:exception: ambiguous iteration
      Iterated regexp: /(([^:\n]|\\\\:)+|[^:\n]*(\\"[^\n]+\\"[^:\n]*)+)(:|:[ \t]*\\\\\n[ \t]*:)/
      '\\:\\:\\::' can be split into
      '\\:|=|\\:\\::'

     and
      '\\:\\:\\::|=|'

    Iterated lens: ./getcap.aug:33.19-.61:

```

Any advice on cleaning that up would be appreciated.

I've tried writing some basic tests however I'm getting stuck where there should be a literal `\` in the output. If I try `\`-escaping it then I get something like the following:

```
tests/test_getcap.aug:102:17: Unexpected character #
tests/test_getcap.aug:101.17-102.8:syntax error, unexpected LIDENT, expecting '}'
tests/test_getcap.aug:syntax error
```

Any ideas?

### DIFF
--- a/lenses/tests/test_getcap.aug
+++ b/lenses/tests/test_getcap.aug
@@ -1,0 +1,122 @@
+module Test_getcap =
+
+(* Example from getcap(3) *)
+let getcap = "example|an example of binding multiple values to names:\\
+     :foo%bar:foo^blah:foo@:\\
+     :abc%xyz:abc^frap:abc$@:\\
+     :tc=more:
+"
+
+test Getcap.lns get getcap =
+  { "record"
+    { "name" = "example" }
+    { "name" = "an example of binding multiple values to names" }
+    { "capability" = "foo%bar" }
+    { "capability" = "foo^blah" }
+    { "capability" = "foo@" }
+    { "capability" = "abc%xyz" }
+    { "capability" = "abc^frap" }
+    { "capability" = "abc$@" }
+    { "capability" = "tc=more" }
+  }
+
+(* Taken from the standard /etc/login.conf *)
+let login_conf = "# Default allowed authentication styles
+auth-defaults:auth=passwd,skey:
+
+# Default allowed authentication styles for authentication type ftp
+auth-ftp-defaults:auth-ftp=passwd:
+
+#
+# The default values
+# To alter the default authentication types change the line:
+#       :tc=auth-defaults:\\
+# to be read something like: (enables passwd, \"myauth\", and activ)
+#       :auth=passwd,myauth,activ:\\
+# Any value changed in the daemon class should be reset in default
+# class.
+#
+default:\\
+        :path=/usr/bin /bin /usr/sbin /sbin /usr/X11R6/bin /usr/local/bin /usr/local/sbin:\\
+        :umask=022:\\
+        :datasize-max=512M:\\
+        :datasize-cur=512M:\\
+        :maxproc-max=256:\\
+        :maxproc-cur=128:\\
+        :openfiles-cur=512:\\
+        :stacksize-cur=4M:\\
+        :localcipher=blowfish,8:\\
+        :ypcipher=old:\\
+        :tc=auth-defaults:\\
+        :tc=auth-ftp-defaults:
+"
+
+test Getcap.lns get login_conf =
+  { "#comment" = "Default allowed authentication styles" }
+  { "record"
+    { "name" = "auth-defaults" }
+    { "capability" = "auth=passwd,skey" }
+  }
+  {  }
+  { "#comment" = "Default allowed authentication styles for authentication type ftp" }
+  { "record"
+    { "name" = "auth-ftp-defaults" }
+    { "capability" = "auth-ftp=passwd" }
+  }
+  {  }
+  {  }
+  { "#comment" = "The default values" }
+  { "#comment" = "To alter the default authentication types change the line:" }
+  { "#comment" = ":tc=auth-defaults:\\" }
+  { "#comment" = "to be read something like: (enables passwd, \"myauth\", and activ)" }
+  { "#comment" = ":auth=passwd,myauth,activ:\\" }
+  { "#comment" = "Any value changed in the daemon class should be reset in default" }
+  { "#comment" = "class." }
+  {  }
+  { "record"
+    { "name" = "default" }
+    { "capability" = "path=/usr/bin /bin /usr/sbin /sbin /usr/X11R6/bin /usr/local/bin /usr/local/sbin" }
+    { "capability" = "umask=022" }
+    { "capability" = "datasize-max=512M" }
+    { "capability" = "datasize-cur=512M" }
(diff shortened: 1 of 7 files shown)


bodgit:
One annoyance I have is that new records will be added without any whitespace like so:
```
foo:bar:baz:
```
which is fine for small records, but long ones will wrap and look a bit meh so I made the following change to my lens:
```diff
diff --git a/lenses/getcap.aug b/lenses/getcap.aug
index bd002de..c38893a 100644
--- a/lenses/getcap.aug
+++ b/lenses/getcap.aug
@@ -28,10 +28,9 @@ module Getcap =
   (* field must not contain ':' unless quoted or '\'-escaped  *)
   let field      = /([^:\n]|\\\\:)+|[^:\n]*(\"[^\n]+\"[^:\n]*)+/
 
-  let sep        = del /:|:[ \t]*\\\\\n[ \t]*:/ ":"
-  let name       = store field . sep
-  let capability = [ label "capability" . store field . sep ]
-  let record     = [ seq "record" . name . capability+ . eol ]
+  let sep        = del /:([ \t]*\\\\\n[ \t]*:)?/ ":\\\n\t:"
+  let capability = [ label "capability" . store field ]
+  let record     = [ seq "record" . store field . sep . capability . ( sep . capability )* . Sep.colon . eol ]
 
   let lns = ( empty | comment | record )*
 
```
which should in theory should mean new records are added like so:
```
foo:\
        :bar:\
        :baz:
```
however what gets written to the file with `augtool` is:
```
foo:
        :bar:
        :baz:
```
The `\`'s aren't written. This and the problem with my tests suggests something weird going on with augeas and literal `\` characters.

lenses/tests/test_getcap.aug
```diff
@@ -0,0 +1,288 @@
...
+        :Fa=\E[34;2~:Fb=\E[35;2~:Fc=\E[36;2~:\
+        :S6=USR_TERM\:vt420pcdos\::k1=\E[11~:k2=\E[12~:\
+        :k3=\E[13~:k4=\E[14~:k5=\E[15~:k6=\E[17~:k7=\E[18~:\
+        :k8=\E[19~:k9=\E[20~:k;=\E[21~:kD=\177:kh=\E[H:\
+        :..px=\EP1;1|%?%{16}%p1%>%t%{0}%e%{21}%p1%>%t%{1}%e%{25}%p1%>%t%{2}%e%{27}%p1%>%t%{3}%e%{30}%p1%>%t%{4}%e%{5}%;%p1%+%d/%p2%s\E\\:\
+        :tc=vt420:
+"
+
+test Getcap.lns get getcap =
+  { "1" = "example|an example of binding multiple values to names"
```
raphink: This test doesn't match what your lens is doing.

raphink:
I've added a CSV lens this morning, which is very similar to what you're trying to achieve.

In fact, `CSV.lns_generic ":"` parses your examples (although not the way you want them parsed).

You might want to check it out.

bodgit:
I realised the structure I was parsing records into didn't allow me easily to write idempotent changes for use with Puppet so I've changed how records are parsed.

I still have a problem with my tests in that I can't give some sample test output containing literal `\` characters and then test that they are present in the parsed structures.

I will take a look at the CVS lens and see how that works.

raphink:
Any news on this lens @bodgit ?

xaque208:
Oooh yes, this would be quite useful.

bodgit:
Still not working 100% but now we know there are some bugs with backslashes in general it makes sense to get those ironed out first.

raphink:
@bodgit typechecking doesn't pass yet though

lutter:
@bodgit nice work .. one minor nit: I now consider `seq` a design mistake since it just confuses people when they try to add to the tree. Would you object to changing `seq "record"` to `label "record"` ? If you don't have time to make that change, I'd be more than happy to do that after merging this PR.

bodgit:
Ironically I think it was `label "record"` originally before I noticed some other lenses use `seq` and figured it might make things more usable. I don't mind either way, I will change it back to use `label` again and update the tests to match.

bodgit:
Other than that, the only thing to point out is that I've dropped reading the actual `termcap` file. Two reasons:

1. It doesn't seem to follow its own conventions w.r.t. escape sequences. AFAICT capabilities should not contain literal `:` characters anywhere and should either be replaced with `\c`/`\C` or its octal value `\072`, I found lots of entries that try to use `\:`. `^` and `\` should also be escaped as `\^` and `\\` and often they're not. This is partly due to `^X` being used as an escape sequence to represent control-X (FSVO X). If I make the lens follow the escape sequence rules then it doesn't parse `termcap`, if I make it aware of the bugs then the lens becomes mostly impossible to pass type checking as there's no consistency.
2. `rtadvd.conf` despite using the same file format doesn't actually use the same library routines for reading the file, hence it contains bare IPv6 addresses, complete with literal `:` characters. Trying to parse that with `getcap` will split `addr="2001:dead::beef"` into `addr="2001`, `dead`, ``, and `beef"` which isn't terribly useful from an Augeas users' PoV.

In theory I should just be able to use a simple regex such as `/[^:]*/` but in order to cater for the most useful files I've used something similar to `/("[^"]*"|[^:"]*)/` so it matches either a quoted string containing anything, or a string containing neither colons nor quotes which seems to work for anything except the `termcap` file.

raphink:
@lutter `seq` is still useful so long as https://github.com/hercules-team/augeas/pull/244 is not merged, or am I missing something?

bodgit:
Actually, thinking about this, I can split this into two lenses with one lens written mostly in terms of the other with just a different regexp for matching the capabilities. Bear with me :smile:

lutter:
@bodgit sounds good .. awaiting the refactored, much expanded lens anxiously ;) Looks like you've stumbled on one of the great things about config files: the discrepancy of the documented format and the format tools actually understand; in a lot of cases, I've had to read the parsing code of whatever consumes a file to discover that. Seems termcap is no different in that regard.

@raphink yes, #244 would still be useful, but I'd like to get a little more feedback on syntax etc. So if you have thoughts, would love to discuss more on the PR

bodgit:
The termcap lens still doesn't work on an OpenBSD `/etc/termcap` but I think that's a bug with their `tic` which generates it; it doesn't escape bare `^` characters whereas if I use `tic` on a RHEL7 host then it generates a file with `\136` instead which the lens parses fine. I try and raise that with them at some point.

lutter:
Awesome ! I am merging this for now; when you resolve the `tic` issues, we can always amend/improve the lens.

PR #277, opened 2015-08-05
Add support for distribution

kumy:
We had this patch here since a long time just waiting to be reported upstream...


raphink:
LGTM :smile:

raphink:
Merged as 8d1e88b

PR #278, opened 2015-08-06
openvpn: added all options available in OpenVPN2.3 and all tests

dafugg:
I've added all possible options in OpenVPN 2.3 from the man page at https://community.openvpn.net/openvpn/wiki/Openvpn23ManPage and tests for lots (all?) permutations of those options.

### DIFF
--- a/lenses/openvpn.aug
+++ b/lenses/openvpn.aug
@@ -1,7 +1,11 @@
 (* OpenVPN module for Augeas
  Author: Raphael Pinson <raphink@gmail.com>
+ Author: Justin Akers <dafugg@gmail.com>
 
  Reference: http://openvpn.net/index.php/documentation/howto.html
+ Reference: https://community.openvpn.net/openvpn/wiki/Openvpn23ManPage
+
+ TODO: Inline file support
 *)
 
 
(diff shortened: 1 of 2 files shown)


dafugg:
I should mention.. the IPv6 options are poorly documented. I've done my best to guess what they should look like but I haven't been able to test them here.

I also didn't add any of the windows-specific options.

dafugg:
I manually ran augparse as the gnulib tests weren't working on my system (test-lock and test-thread_create fail in gnulib). I've run just the augeas lense tests and openvpn passes here:
PASS: lens-openvpn.sh
however the check takes quite a while. Could someone else confirm that the test passes?

raphink:
```
$ \time -v augparse -I lenses/ lenses/openvpn.aug

	Maximum resident set size (kbytes): 4266356
```

is the problem

Just switching to `Rx.ip` instead of `Rx.ipv4` already eats up almost 2GB, because of the complexity of `Rx.ipv6`. I think it would be reasonable to make this lens a bit less strict on values, to save a bit of memory on typechecking it.

raphink:
Replacing all occurrences of `Rx.ipv6` with `Rx.word` already gets the max resident size down to 2062396.

Maybe it would be good to have a local, loose lens for IPv6 (something like `/[0-9A-Fa-f:]+/` or so) that would already save 2GB.

dafugg:
Thanks for the update @raphink . I have built augeas here in a Yocto 'ptest' image to run in a VM with a hard 1.5GB limit. All of the tests passed consistently in this environment.

raphink:
LGTM, I'll merge.

Issue #279, opened 2015-08-20
Spacevars lens fails on empty values


nielslaukens:
When using the Spacevars lens, it fails to parse a file where a particular key has no value:
```
key 
# note that their is a trailing space on the line above
```
This happens for example on `nslcd.conf`, which has an empty `uri` key right after installation.

Error output:
```
/error = "parse_failed"
/error/pos = "70"
/error/line = "4"
/error/char = "0"
/error/lens = "/usr/share/augeas/lenses/dist/spacevars.aug:31.10-.32:"
```
(The error position is the line containing a key with an empty value)

I've traced this back through the lenses, and the (eventual) culprit is in [Rx.space_in](https://github.com/hercules-team/augeas/blob/8d1e88ba94adab95fb4d9fe17e5864c94ec11dfb/lenses/rx.aug#L39), which doesn't match an empty string. [The documentation](http://augeas.net/docs/references/lenses/files/rx-aug.html#Rx.space_in) isn't entirely clear on the issue:
> A string which not starting or ending with a space

But the way I understand that, it should match an empty string.

If the current behavior is the intended behavior, this issue can also be fixed along the way by using another lens/regexp:
* [Build.key_ws_value](https://github.com/hercules-team/augeas/blob/8d1e88ba94adab95fb4d9fe17e5864c94ec11dfb/lenses/build.aug#L144)
* [Spacevars.entry](https://github.com/hercules-team/augeas/blob/8d1e88ba94adab95fb4d9fe17e5864c94ec11dfb/lenses/spacevars.aug#L25)

raphink:
As stated in #280, `Rx.space_in` should not be modified to accept an empty string. However, modifying `Spacevars.entry` to accept empty values is an option.

Usually, I try to follow some kind of standard for this:

```
{ "key" }
```

maps to a flag, e.g.

```
key
```

while

```
{ "key" = "" }
```

maps to an empty value, e.g.

```
key =
```

It seems to me that you need support for the second kind, right?

raphink:
Oh wait, it's Spacevars, so there's no "=" anyway, so it doesn't matter either way.

raphink:
If https://github.com/raphink/augeas/commit/cb5fb3066ee89b9143d3181b6fc769d4e8e1518c fixes the issue for you, I'll merge it as a fix for this bug.

nielslaukens:
raphink@cb5fb30 fixes this issue for me. Thank you!

PR #280, opened 2015-08-20
Make Rx.space_in also accept the empty string

nielslaukens:
This fixes issue #279


raphink:
This is not a good idea. Many lenses use things like

```augeas
Sep.space . store Rx.space_in . Sep.space
```

If `Rx.space_in` allows an empty value, then it introduces concatenation ambiguities in tons of existing lenses. I'll close this PR and discuss the other possible options in #279.

nielslaukens:
Indeed, it breaks lots of things; I figured as much from the Travis CI output.

PR #281, opened 2015-08-20
Puppetfile.lens fix: modules do not always respect the puppet module dashed naming convention

fconrotte:
As discussed over here:
https://github.com/pzim/reaktor/issues/15#issuecomment-131135811

Some enterprise modules do not always respect the puppet module dashed naming convention


raphink:
The problem is that this patch doesn't pass unit tests. Can you provide a new unit test to illustrate what you're trying to achieve?

raphink:
This seems to have been fixed in 211a56a4ae234a106ff14405e9371288564be762 already.

PR #282, opened 2015-08-22
Rsyslog: Improve property filter parsing.

gasmith-nutanix:
- Treat whitespace after commas as optional.
 - Recognize '~' as a valid syslog action (discard).


raphink:
Thank you.

PR #215, opened 2015-08-25
shellvars lens can't handle "expression1 && expression2" or "expression1 || expression2"

raphink:
The shellvars lens fails to handle code like the following:

```
[ ! -z $DEBUG ] && ENABLE_DEBUG="1"
```

or

```
grep -q "Debian" /etc/issue && DEBIAN=1
```

or other variations of this kind of expressions.


raphink:
@chelli I've just attached some code that fixes the second case.

PR #283, opened 2015-08-25
reprepro_uploader lens doesn't support groups

raphink:
reprepro's `uploaders` file can contain `group` definitions, see [doc](https://mirrorer.alioth.debian.org/reprepro.1.html#UPLOADERS%20FILES). Unless I'm wrong, those items are not parsed by the lens.

```
allow condition by group groupname
   which allows every member of group groupname. Groups can be manipulated by

group groupname add key-id
   to add a key-id (see above for details) to this group, or

group groupname contains groupname
   to add a whole group to a group.
  To avoid warnings in incomplete config files there is also

group groupname empty
   to declare a group has no members (avoids warnings that it is used without those) and

group groupname unused
   to declare that a group is not yet used (avoid warnings that it is not used).
```

### DIFF
--- a/lenses/reprepro_uploaders.aug
+++ b/lenses/reprepro_uploaders.aug
@@ -179,10 +179,13 @@ let allow =
 (* View: group
    A group declaration *)
 let group =
-    [ key "group" . Sep.space
-  . store Rx.word . Sep.space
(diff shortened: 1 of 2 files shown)


raphink:
Here's the code. Please test and let me know if that works.

PR #294, opened 2015-09-08
Dhclient: avoid put ambiguity for node without value

raphink:
When `dhclient.conf` contains a line like:

```
send host-name = gethostname();
```

*any* change to the file fails to save, since it expects a mandatory value for the node. This PR fixes this by separating the case with a node value from the cases without.


raphink:
@lutter: as a note, `augparse` did not automatically detect this ambiguity, a `put` test was necessary.

PR #295, opened 2015-09-14
Aptsources lens can't handle [ options ] in /etc/apt/sources.list

raphink:
Sources.list files may contain options in the format "[arch=amd64,armhf]" immediately after type field.  This is not understood properly understood by the aptsources lens leading to incorrect parsing.  For example, with the following repository lines in sources.list:

```
deb [arch=amd64] tor+http://ftp.us.debian.org/debian sid main contrib
deb [ arch=amd64 trusted=true ] http://ftp.us.debian.org/debian sid main contrib
```
Augeas parses parses as follows:
```
augtool> ls /files/etc/apt/sources.list/1
type = deb
uri = [arch=amd64]
distribution = tor+http://ftp.us.debian.org/debian
component[1] = sid
component[2] = main
component[3] = contrib
component[4] = non-free
augtool> ls /files/etc/apt/sources.list/2
type = deb
uri = [
distribution = arch=amd64
component[1] = trusted=true
component[2] = ]
component[3] = http://ftp.us.debian.org/debian
component[4] = sid
component[5] = main
component[6] = contrib
component[7] = non-free
```
Options can contain multiple settings as key=value pairs separated by spaces.  Multiple values are possible as a comma separated list.  Current set of options are "arch=", "arch+=", "arch-=" and "trusted=".  For more information about options see man page for sources.list.

Augeas should properly parse and understand options so that not are other fields properly understood but also options and read and set properly.

### DIFF
--- a/lenses/aptsources.aug
+++ b/lenses/aptsources.aug
@@ -10,7 +10,7 @@ module Aptsources =
  * Group: Utility variables/functions
  ************************************************************************)
   (* View:  sep_ws *)
-  let sep_ws = del /[ \t]+/ " "
+  let sep_ws = Sep.space
 
   (* View: eol *)
   let eol = Util.del_str "\n"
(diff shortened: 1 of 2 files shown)


raphink:
Attached commits to fix this.

PR #296, opened 2015-09-14
Aptsources lens can't handle cdrom:// URLs properly

raphink:
CDROM URLs in /etc/apt/sources.list may contain spaces in them.  This is not properly handled by `aptsources` lens properly.  For example, the following is a valid sources.list repository entry.
```
deb cdrom:[Debian GNU/Linux 7.5.0 _Wheezy_ - Official amd64 CD Binary-1 20140426-13:37]/ wheezy main
```
With this entry, aptsources lens parses as follows:
```
augtool> ls /files/etc/apt/sources.list/1
type = deb
uri = cdrom:[Debian
distribution = GNU/Linux
component[1] = 7.5.0
component[2] = _Wheezy_
component[3] = -
component[4] = Official
component[5] = amd64
component[6] = CD
component[7] = Binary-1
component[8] = 20140426-13:37]/
component[9] = wheezy
component[10] = main
augtool> 
```

Aptsources lens should be fixed to allow spaces in URL when enclosed in "[" and "]".

### DIFF
--- a/lenses/aptsources.aug
+++ b/lenses/aptsources.aug
@@ -23,6 +23,14 @@ module Aptsources =
   (* View: word *)
   let word = /[^][# \n\t]+/
 
+  (* View: uri *)
+  let uri =
+       let protocol = /[a-z+]+:/
+    in let path = /\/[^] \t]*/
+    in let path_brack = /\[[^]]+\]\/?/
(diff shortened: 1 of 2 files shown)


raphink:
Attached commit to fix this.

PR #300, opened 2015-09-16
WIP: Lua interpreter in augtool

raphink:
This PR adds a `-l`/`--lua` flag to augtool which allows to code in Lua instead of the native augrun interpreter.

Examples can be found in `tests/test.auglua`:

```lua
--- test aug_get()
assert(aug_get("/augeas/version") == "1.4.0")

--- test aug_label()
assert(aug_label("etc/passwd/rpinson") == "rpinson")

--- test aug_set()
aug_set("/foo", "bar")
assert(aug_get("/foo") == "bar")

--- test aug_setm()
aug_setm("etc/passwd/*", "shell", "zsh")
assert(aug_get("etc/passwd/rpinson/shell") == "zsh")

--- test aug_span()
--- TODO

--- test aug_insert()
aug_insert("etc/passwd/*[1]", "foo", true)
assert(aug_label("etc/passwd/*[1]") == "foo")

--- test aug_rm()
aug_rm("etc/passwd/rpinson")
assert(aug_label("etc/passwd/rpinson") == nil)

--- test aug_mv()
aug_mv("etc/passwd/root", "etc/passwd/qux")
assert(aug_label("etc/passwd/*[last()]") == "qux")
assert(aug_label("etc/passwd/root") == nil)

--- test aug_cp()
aug_cp("etc/passwd/qux", "etc/passwd/root")
assert(aug_get("etc/passwd/root/uid") == aug_get("etc/passwd/qux/uid"))

--- test aug_rename()
aug_rename("etc/passwd/qux", "bar")
assert(aug_get("etc/passwd/root/uid") == aug_get("etc/passwd/bar/uid"))

--- test aug_match()
m = aug_match("etc/passwd/*")
c = 0
for k, v in pairs(m) do c= c+1 end
assert(c > 0)
assert(m[1] == "/files/etc/passwd/foo")

--- test aug_defvar()
aug_defvar("root", "etc/passwd/root")
assert(aug_label("$root") == "root")

--- test aug_defnode()
aug_defnode("raphink", "etc/passwd/raphink/uid", "314")
assert(aug_get("$raphink") == "314")
--- TODO: support 3 arguments
--- aug_defnode("gid", "etc/passwd/raphink/gid")
--- assert(aug_get("$gid") == nil)

--- test aug_save()
--- expect error here!
aug_save()

--- test aug_load()
aug_load()

--- test aug_text_store()
aug_set("/input", "key=value\n")
aug_text_store("Shellvars.lns", "/input", "/parsed")
assert(aug_get("/parsed/key") == "value")

--- test aug_text_retrieve()
aug_set("/parsed/key", "newval")
aug_text_retrieve("Shellvars.lns", "/input", "/parsed", "/output")
assert(aug_get("/output") == "key=newval\n")

--- test aug_escape_name()
--- test that?

--- test aug_transform()
aug_transform("Json.lns", "/tmp/foo.json", false)
assert(aug_get("/augeas/load/Json/incl[1]") == "/tmp/foo.json")

--- test aug_print()

--- test aug_to_xml()

--- test aug_srun()

--- test aug_errors()
```


raphink:
I kind of fixed the multiline lua REPL problem, irb-style. The implementation might not be perfect, but it works:

```
augtool|lua> for k, v in pairs(match("etc/passwd/*")) do
augtool|lua?>   print(aug_get(v .. "/uid"))
augtool|lua?>   end
0
1
2
3
4
5
6
7
8
10
11
12
13
14
99
69
32
29
4294967294
```

raphink:
@ncopa since you wrote the bindings for lua, do you have an opinion on this?

ncopa:
what is the benefit with embedding lua into augeas instead of just using lua-augeas module?

raphink:
@ncopa most users (at least sysadmins) will not want to install a lua REPL, the lua bindings for Augeas, and write a script that sets up a handler (if they know how that works), but they need to be able to perform operations and loop in the code they use in augtool

The idea here is that you can simply install `augeas-tools`, launch `augtool -l` and you have an already set up `aug` library, and you can use the augtool flags to control the fake root, lens lib dir, etc.

The typical use I'm thinking of is setting up docker containers using augtool-based scripts.

lutter:
I really like the idea of having some sort of standard scripting language readily available when you install augeas - and lua seems to be a great fit for that. This is awesome stuff.

I am not 100% sold on whether that should be part of augtool or a separate tool - I need to understand the tradeoffs better in terms of platform suppport, addl dependencies, and (maybe) size. I am not against it, just that I need a clearer picture of these things.

Also: if we ship with lua support by default, wouldn't we also want some sort of basic lua scripts/libraries, i.e. the ability to ship conveniences in lua, basically things that either make working with the augeas API easier from lua, or that give you canned answers for common tasks ? That doesn't have to happen right off the bat, but is something to keep in mind as this evolves.

raphink:
Thanks for reviewing @lutter.

From what I can see, Lua supports at least as many platforms as we do (Unix + Windows). It is known for being not only embeddable, but also very light. I've tried building Augeas statically to compare, but there is a local definition of `xasprintf` which conflicts when building statically.

As for external Lua libraries, I believe you should be able to use `require 'somelib'` in augtool with the current code.

raphink:
I commented out the definition of `xasprintf` in `augtool.c` and managed to build statically (or so it seems). So here are the sizes of the augtool binary:

* without lua support: 1584769
* with lua support: 1618371

So Lua support adds 33602 (32K) to the static binary.

src/augeas_sym.version
```diff
@@ -69,4 +69,6 @@ AUGEAS_0.18.0 {
 AUGEAS_0.20.0 {
     global:
       aug_escape_name;
+      luaopen_augeas;
+      aug_lua;
```
igalic: has there been no release since your patch?
i.e.: shouldn't this be 0.21.0?

src/auglua.c
```diff
@@ -0,0 +1,531 @@
...
+  path = luaL_checkstring(L, 1);
+
+  augeas *aug = checkaug(L);
+  r = aug_get(aug, path, &value);
+  if (r < 0)
+      return lua_pusherror(L);
+  lua_pushstring(L, value);
+
+  /* return the number of results */
+  return 1;
```
igalic: this comment seems misleading

src/auglua.c
```diff
@@ -0,0 +1,531 @@
...
+  // TODO: check string really
+
+  augeas *aug = checkaug(L);
+  r = aug_label(aug, path, &value);
+  if (r < 0)
+      return lua_pusherror(L);
+  lua_pushstring(L, value);
+
+  /* return the number of results */
+  return 1;
```
igalic: again, misleading comment.

src/auglua.h
```diff
@@ -0,0 +1,48 @@
...
+ * Author: Raphaël Pinson <raphael.pinson@camptocamp.com>
+ */
+#include <lua.h>
+
+#ifndef AUGLUA_H_
+#define AUGLUA_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
igalic: is the rest of the project written in C++, or why are you adding this protective `extern`?

src/augtool.c
```diff
@@ -481,7 +534,14 @@ static int main_loop(void) {
 
     while(1) {
         if (get_line) {
```
igalic: following the logic, and the variables' malloc/free/assignment is really hard in this loop :(
how hard would it be to put all of it in very small (inlined?) functions?

src/augeas_sym.version
raphink: You're right, that was a testing thing. I guess this entry also depends on whether we expose aug_lua

src/auglua.c
raphink: Iirc I copied it from the lua doc. I could remove it indeed.

src/auglua.h
raphink: The rest of augeas has that, I suppose to allow using the lib in c++

src/augtool.c
raphink: Yes, it might be a good idea to refactor that

igalic: you might wanna take a shot at refactoring it while you're extending it, because… well, you're adding more glue code onto it, and it's already hard to follow

raphink: Maybe it would be easier to refactor before in fact, so as to make this PR more simple.

ncopa:
I would like to merge the majority of this with https://github.com/ncopa/lua-augeas/blob/master/laugeas.c into a common C file. Then we could have a separate make target for a lua module, so we could build the lua-augeas directly from the augeas sources, in addition to the lua embedded stuff.

src/auglua.h
lutter: The functions in `auglua.h` aren't meant to become part of the public augeas API, or are they ? I.e., when would somebody else write `include "auglua.h"` in their source ?

src/augtool.c
lutter: What's the value to the user that they can type `augtool -l` rather than `auglua` to get the lua-based shell ? With the latter, `auglua` would become a separate executable.

raphink: Yes, although `auglua` would need to take all the same options are `augtool`.

src/auglua.h
raphink: Would it make sense to expose an `aug_lua` call, which would allow e.g. Puppet to take lua commands in the `augeas` type?

raphink:
@ncopa which parts exactly would you like to merge into lua-augeas?

raphink:
Rebased on master

kunkku:
@raphink I understand ncopa's comment such that there is a considerable overlap between your auglua.c and his laugeas.c. He would like to abolish his lua-augeas project and build the Lua bindings from the Augeas source tree in future.

raphink:
I understand that... I don't have much time for that (or extended knowledge of Lua), but I have several projects that would benefit from lua integration in augtool...

raphink:
@ncopa do you have some time maybe to help me implement that properly?

thedrow:
Will this introduce another dependency on libaugeas?

kunkku:
Yes, libaugeas would depend on a specific version of the Lua interpreter. This might cause dynamic symbol conflicts if one attempts to use lua-augeas with a different interpreter version. Even minor Lua releases often introduce backwards incompatible changes, so we should be flexible with the versions.

@raphink What do you think of moving the Lua extensions to a separate library, e.g. libauglua?

raphink:
@kunkku there is already a lua-augeas project, but the goal here is really to embed lua in augtool.

raphink:
@kunkku just added commits to this PR (thank you).

lutter:
I've recently started playing with [mruby](http://mruby.org) and really like it. I ported the ruby bindings over [to it](https://github.com/hercules-team/mruby-augeas)

All of this makes me wonder if an mruby-based command line tool might not be a better option, as it seems a lot more people are familiar with that than with Lua. To give you some impression of what that looks like, [this](https://github.com/lutter/libral/blob/d732a02b356cdf48d78b3d588c9b832b5a916f00/data/providers/ssh_authorized_key.prov) is an example of using mruby with augeas

raphink:
I've seen lua in a lot of projects, but never mruby, so i can't really tell. I guess all that matters would be to bring higher level language into augeas itself.

PR #303, opened 2015-09-28
Improve performance of some path expressions

lutter:
Path expressions of the form `service[75]` were very slow when there were a large number of `service` nodes. These patches try to speed that up.

Tons of thanks to @domcleal for figuring out why this was slow, writing a test that demonstrates the problem and providing lots of detail on what exactly was going on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,12 +35,14 @@ src/parser.[ch]
 src/lexer.[ch]
 src/augtool
 src/augparse
+src/callgrind.*
 tests/fatest
 tests/leak
 tests/lens-*.sh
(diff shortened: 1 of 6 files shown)


domcleal:
That's excellent, thanks for taking this up.  The optimisations all make sense, and work very well with the test case (68064ms, 18069ms, 1139ms, 318ms per respective commit).  Even without the special case it's such a significant improvement that it should be usable for other predicates.

:+1:

lutter:
Thanks for all the work in figuring out where the perf bottlenecks are - that made writing this a ton easier !

PR #309, opened 2015-11-04
Fix a Clang 'sometimes-uninitialized' warning

dtrebbien:
clang-700.1.76 was issuing a 'sometimes-uninitialized' warning for
minimize_hopcroft() because it was possible for control flow to reach
the \`for (int i=0; i < nstates*nsigma; i++)' loop in done: before
\`nstates' and/or \`nsigma' were initialized.


lutter:
Good catch !

Rebased and merged as 8d223753

PR #311, opened 2015-11-04
Fix a Clang 'sometimes-uninitialized' warning in aug_escape_name()

dtrebbien:
If \`in' or \`out' is NULL, then the uninitialized value of \`result' was
returned.


lutter:
Good catch !

Rebased and merged as 3458d8ba

PR #313, opened 2015-11-05
Libfa const correctness

dtrebbien:


src/fa.c
```diff
@@ -437,6 +437,10 @@ static struct state *add_state(struct fa *fa, int accept) {
     for (struct trans *t = (s)->trans;                                  \
          (t - (s)->trans) < (s)->tused;                                 \
          t++)
+#define for_each_const_trans(t, s)                                      \
+    for (const struct trans *t = (s)->trans;                            \
```
lutter: Another way to do this is to just have `for_each_trans` but use `typeof ((s)->trans) t`; or would that break clang ?

src/fa.c
```diff
@@ -2217,8 +2227,12 @@ struct fa *fa_intersect(struct fa *fa1, struct fa *fa2) {
...
 
     if (fa1->nocase != fa2->nocase) {
-        F(case_expand(fa1));
-        F(case_expand(fa2));
+        cfa1 = fa_clone(fa1);
+        cfa2 = fa_clone(fa2);
+        if (cfa1 == NULL || cfa2 == NULL)
+            goto error;
+        F(case_expand(cfa1));
+        F(case_expand(cfa2));
```
lutter: Couldn't we now do `fa1 = cfa1` so that we can avoid all the places where we have `(cfa1 ? cfa1 : fa1)` ?

src/fa.c
```diff
@@ -2283,6 +2300,8 @@ struct fa *fa_intersect(struct fa *fa1, struct fa *fa2) {
             fa = NULL;
         }
     }
+    fa_free(cfa1);
+    fa_free(cfa2);
```
lutter: Yikes .. thanks for fixing this leak

src/fa.c
```diff
@@ -2217,8 +2227,12 @@ struct fa *fa_intersect(struct fa *fa1, struct fa *fa2) {
         return fa_make_empty();
 
     if (fa1->nocase != fa2->nocase) {
-        F(case_expand(fa1));
-        F(case_expand(fa2));
+        cfa1 = fa_clone(fa1);
+        cfa2 = fa_clone(fa2);
```
lutter: Actually, I don't really like this; if you use the same FA repeatedly for intersection testing, you incur the cost for cloning every time. Before, that was a onetime cost.

lutter:
While reviewing this I actually remembered why these functions are not const-correct: the API contract is that any FA you pass in will continue to describe the same regular language, but we might change the actual automaton behind your back (e.g., determinize it).

Since some of these changes, like determinization, can be quite expensive I'd rather not merge this patch to have the functions be const everywhere.

Are the current non-const prototypes causing problems somewhere ?

dtrebbien:
> While reviewing this I actually remembered why these functions are not const-correct: the API contract is that any FA you pass in will continue to describe the same regular language, but we might change the actual automaton behind your back (e.g., determinize it).

Okay, that makes sense.

> Are the current non-const prototypes causing problems somewhere ?

No, not really.  I was working on a [Rust FFI wrapper for libfa](https://github.com/dtrebbien/fa-rs) and I wanted to figure out whether these functions modified the data structure in any way for the purpose of figuring out mut/const semantics on the Rust APIs.

Now that I understand the design of allowing the automaton to be changed as long as it represents the same regular language, I agree that this shouldn't be merged.  There is, however, a bug fix among these commits ([6e7d090](https://github.com/dtrebbien/augeas/commit/6e7d090dcdddf4147ccc71a796c4c1858a3fe007)) and some minor documentation improvements that would be good to apply.  I will close this PR and open another for those changes.

dtrebbien:
I have opened PRs #321 and #322.

lutter:
Thanks for splitting these out; I'll have a look in a minute.

I am very happy that you are writing Rust bindings for libfa; you probably know, and it will likely not help much, but there are also bindings for augeas: https://github.com/panicbit/rust-augeas/commits/master It may or may not make sense to try and combine efforts.

dtrebbien:
I wasn't aware of rust-augeas.  Thanks for mentioning it.  I will definitely look into this other project.

Issue #307, opened 2015-10-29
httpd lens fails on backslashes in directive regular expressions


erebus2:
Augeas currently throws “syntax errors” when an apache config happens to include regular expressions.

A regular expressions is not a syntax error. In fact, regexes are quite common in apache configs as they are needed for mod_rewrite.

This issue is happening with the httpd lens from augeas v1.4.

See also: https://github.com/letsencrypt/letsencrypt/issues/1202

domcleal:
Could you provide part of a config file that fails please?

pde:
There's a sample in this ticket:

https://github.com/letsencrypt/letsencrypt/issues/1264

domcleal:
@pde I don't see a regular expression there.  The config file in that ticket works fine on Augeas 1.4.0.

pde:
I suspect these may have been the samples we were looking for:

```
#
# Apache/PHP/Drupal settings:
#

# Protect files and directories from prying eyes.
<FilesMatch "\.(engine|inc|info|install|make|module|profile|test|po|sh|.*sql|theme|tpl(\.php)?|xtmpl)(~|\.sw[op]|\.bak|\.orig|\.save)?$|^(\..*|Entries.*|Repository|Root|Tag|Template)$|^#.*#$|\.php(~|\.sw[op]|\.bak|\.orig\.save)$">
  Order allow,deny
</FilesMatch>

# Don't show directory listings for URLs which map to a directory.
Options -Indexes

# Follow symbolic links in this directory.
Options +FollowSymLinks

# Make Drupal handle any 404 errors.
ErrorDocument 404 /index.php

# Set the default handler.
DirectoryIndex index.php index.html index.htm
```

```
# Apache config file for anarcat.wiki.orangeseeds.org
# Automatically generated by ikisite; do not modify directly.

<VirtualHost *:80>
        ServerAdmin root@localhost
        ServerName anarcat.wiki.orangeseeds.org:80


        SuexecUserGroup w-anarcat w-anarcat


        UserDir disabled

        RewriteEngine On
        RewriteRule ^/(.*) http\:\/\/anarc\.at\/$1 [L,R,NE]

        ErrorLog /var/log/ikiwiki-hosting/w-anarcat/error.log
        LogLevel warn
        CustomLog /var/log/ikiwiki-hosting/w-anarcat/access.log combined
</VirtualHost>
```

pde:
[...]

PR #318, opened 2015-11-24
Httpd: parse backslashes in unquoted directive args

domcleal:
Fixes issue #307

### DIFF
--- a/lenses/httpd.aug
+++ b/lenses/httpd.aug
@@ -59,7 +59,7 @@ let empty               = Util.empty_dos
 let indent              = Util.indent
 
 (* borrowed from shellvars.aug *)
-let char_arg_dir  = /[^\\ '"\t\r\n]|\\\\"|\\\\'/
+let char_arg_dir  = /([^\\ '"\t\r\n]|[^\\ '"\t\r\n][^ '"\t\r\n]*[^\\ '"\t\r\n])|\\\\"|\\\\'/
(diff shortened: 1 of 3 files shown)


raphink:
LGTM

lenses/httpd.aug
```diff
@@ -59,7 +59,7 @@ let empty               = Util.empty_dos
 let indent              = Util.indent
 
 (* borrowed from shellvars.aug *)
-let char_arg_dir  = /[^\\ '"\t\r\n]|\\\\"|\\\\'/
+let char_arg_dir  = /([^\\ '"\t\r\n]|[^\\ '"\t\r\n][^ '"\t\r\n]*[^\\ '"\t\r\n])|\\\\"|\\\\'/
```
pde: I can't review this for correctness yet but I'm going to try to describe what this changes in English.  

Previously, change_arg_dir could be any single non(backslash or "terminator") , or an escaped single or double quote.

Now change_arg_dir could be those things, or any string of two or more non"terminator" chacters, with backslashes permitted in any position except the first and last in the string.

"terminator" means whitespace or quote.

domcleal: Yep, that's a correct assessment.

domcleal:
Cheers, merged as 155746c.

pde:
Thanks for turning this around so quickly @domcleal!  I think I found a case that this patch didn't catch, and have opened #324 for that.

Issue #319, opened 2015-11-24
double free when deleting nested XML elements using // (descendant-or-self)


GeoffWilliams:
## Overview
Double free from libaugeas with lens `Xml.lns` when deleting nested XML elements collected using `//` (descendant-or-self).

## Software versions
* CentOS Linux release 7.0.1406 (Core) x86_64
* Augeas 495358aad67bf51eac1019f4ac20c2437c52afbe (Wed Nov 18 11:22:44 2015)
* Ruby ruby 2.1.7p400 (2015-08-18 revision 51632) [x86_64-linux]
* ruby-augeas (0.5.0)

## background
Augeas is being used to remove all `Realm` elements from Tomcat's `server.xml` file using the `ruby-augeas` API like this:
```ruby
@aug.rm("/files/opt/apache-tomcat/tomcat6/conf/server.xml/Server//Realm")
```
The operation itself seems to succeed but there is an error from `libaugeas` when the library attempts to free the memory being used.

Originally this problem was encountered invoking augeas through Puppet but I was able to reproduce the problem using `augtool`, see testcase.

External bug report:  https://tickets.puppetlabs.com/browse/MODULES-1721

## Error details
The above Ruby function enters libaugeas in `aug_rm()` and results in the following backtrace when running under puppet:
```gdb
#0  __lll_lock_wait_private () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:95
#1  0x00007f0edd206f66 in _L_lock_12192 () at malloc.c:5200
#2  0x00007f0edd2044e1 in __GI___libc_malloc (bytes=56) at malloc.c:2866
#3  0x00007f0ede388388 in _dl_map_object_deps (map=map@entry=0x4593280,
    preloads=preloads@entry=0x0, npreloads=npreloads@entry=0, trace_mode=trace_mode@entry=0,
    open_mode=open_mode@entry=-2147483648) at dl-deps.c:515
#4  0x00007f0ede38e92c in dl_open_worker (a=a@entry=0x12f2408) at dl-open.c:265
#5  0x00007f0ede38a314 in _dl_catch_error (objname=objname@entry=0x12f23f8,
    errstring=errstring@entry=0x12f2400, mallocedp=mallocedp@entry=0x12f23f0,
    operate=operate@entry=0x7f0ede38e7b0 <dl_open_worker>, args=args@entry=0x12f2408)
    at dl-error.c:177
[...]

PR #320, opened 2015-11-24
aug_rm: fix segfault when deleting a tree and one of its ancestors

lutter:
In a tree like /files/1/2, when we execute 'rm /files//*', the path
expression matches /files/1 and /files/1/2. When tree_rm goes to delete
these two nodes, it first deletes (frees) /files/1 and all its
descendents. By the time we try to delete /files/1/2, the pointer we have
to that is no longer valid and we end up causing a double-free.

With this change, we make sure we only delete a node if none of its
ancestors is being deleted beforehand in the same operation - deleting a
node, and one of its ancestors afterwards is fine as the pointer to the
ancestor is still valid.

Fixes https://github.com/hercules-team/augeas/issues/319

Special shoutout to Geoff Williams for finding, diagnosing and filing a
great bug report about this.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -1142,6 +1142,7 @@ int tree_rm(struct pathx *p) {
     struct tree *tree, **del;
     int cnt = 0, ndel = 0, i;
 
+    /* set ndel to the number of trees we could possibly delete */
     for (tree = pathx_first(p); tree != NULL; tree = pathx_next(p)) {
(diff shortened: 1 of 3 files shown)


GeoffWilliams:
Wow that was quick, thanks @lutter !  

I was originally drafting the report as a request for help fixing this and in the process identified the root cause of the bug.  I've tested the PR and can confirm this fixes the problem I was having.

Thanks!
Geoff

PR #321, opened 2015-11-25
Fix fa_ambig_example() when the automata involve NULs

dtrebbien:
After the example for \`amb' was calculated, the code that generated the
*UPV string and pointers *PV and *V within this string used strlen() to
calculate the length of the example; however, the example string may
contain NUL bytes. Thus, it could happen that not enough space would be
allocated for the *UPV string.

Instead of computing the length of the example using strlen(), just use
\`s_len'.

Added a new test to fatest, testAmbigWithNuls.


lutter:
Merged as 8a292c1b

PR #325, opened 2015-11-27
Httpd: parse backslashes at the start of directive args

### LINKED ISSUE #324: Escaped dots in regexps crashing the httpd lens
I still see augeas parser crashes with this file:

https://github.com/letsencrypt/letsencrypt/files/36394/htaccess.txt

domcleal:
Fixes #324


domcleal:
Merged as cb85b59.

PR #327, opened 2015-12-21
httpd lens fails on parsing perl blocks

raphink:
a perl section like
	<Perl>
		Apache::AuthDBI->setCacheTime(600);
	</Perl>
in a VirtualHost gives a parse_failed error for the file which defines the vhost

Errror:
/augeas/files/etc/apache2/sites-available/000-test.conf/error = "parse_failed"
/augeas/files/etc/apache2/sites-available/000-test.conf/error/pos = "108"
/augeas/files/etc/apache2/sites-available/000-test.conf/error/line = "7"
/augeas/files/etc/apache2/sites-available/000-test.conf/error/char = "2"
/augeas/files/etc/apache2/sites-available/000-test.conf/error/lens = "/usr/share/augeas/lenses/dist/httpd.aug:98.10-.44:"
/augeas/files/etc/apache2/sites-available/000-test.conf/error/message = "Syntax error"

 sample vhost file:
<VirtualHost *:80>
	ServerAdmin webmaster@muze.nl
	DocumentRoot /var/wwww/
	ServerName localhost

	<Perl>
		Apache::AuthDBI->setCacheTime(600);
	</Perl>

</VirtualHost>

### DIFF
--- a/lenses/httpd.aug
+++ b/lenses/httpd.aug
@@ -95,10 +95,19 @@ let arg_wordlist =
 
 let argv (l:lens) = l . (sep_spc . l)*
 
+(* arg_dir_msg may be the last or only argument *)
+let dir_args = (argv (arg_dir|arg_wordlist) . (sep_spc . arg_dir_msg)?) | arg_dir_msg
+
 let directive =
-    (* arg_dir_msg may be the last or only argument *)
-     let dir_args = (argv (arg_dir|arg_wordlist) . (sep_spc . arg_dir_msg)?) | arg_dir_msg
-  in [ indent . label "directive" . store word .  (sep_spc . dir_args)? . eol ]
+  [ indent . label "directive" . store word .  (sep_spc . dir_args)? . eol ]
+
+let perl_directive =
+     let perl_word = word . ("::" . word)* . "->" . word
+  in [ indent . label "perl_directive" . store perl_word
+     . Sep.opt_space . Util.del_str "(" . Sep.opt_space
+     . dir_args
+     . Sep.opt_space . Util.del_str ")" . Sep.opt_space . Util.del_str ";"
+     . eol ]
 
 let section (body:lens) =
     (* opt_eol includes empty lines *)
@@ -110,7 +119,7 @@ let section (body:lens) =
     let dword = del word "a" in
         [ indent . dels "<" . square kword inner dword . del />[ \t\n\r]*/ ">\n" ]
 
-let rec content = section (content|directive)
+let rec content = section (content|directive|perl_directive)
 
 let lns = (content|directive|comment|empty)*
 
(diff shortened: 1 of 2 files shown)


raphink:
I attached a patch to the issue with a possible implementation, mapping to `perl_directive` nodes.

lenses/httpd.aug
```diff
@@ -95,10 +95,19 @@ let arg_wordlist =
...
+let dir_args = (argv (arg_dir|arg_wordlist) . (sep_spc . arg_dir_msg)?) | arg_dir_msg
+
 let directive =
-    (* arg_dir_msg may be the last or only argument *)
-     let dir_args = (argv (arg_dir|arg_wordlist) . (sep_spc . arg_dir_msg)?) | arg_dir_msg
-  in [ indent . label "directive" . store word .  (sep_spc . dir_args)? . eol ]
+  [ indent . label "directive" . store word .  (sep_spc . dir_args)? . eol ]
+
+let perl_directive =
+     let perl_word = word . ("::" . word)* . "->" . word
```
domcleal: I'm not sure Perl should be parsed to this extent, should it just be a free-form value inside `<Perl>`?

raphink: We don't have access to the "inside <Perl>" info in the lens though, unless we make `Perl` a reserved word for tags, and change the directives available inside (which could be a good option).

domcleal: Yes, I was thinking of excluding it from the existing section re and handling it specially.

raphink:
@domcleal there you go, what do you think?

domcleal:
Yeah, that's more like what I had in mind - but would you then remove the parsing of the contents of the Perl script?

raphink:
Ah, right, we could treat the whole string between the `Perl` tags as a blob of Perl code indeed.

domcleal:
Exactly.

raphink:
There you go @domcleal

domcleal:
:+1:

lenses/httpd.aug
```diff
@@ -106,11 +106,17 @@ let section (body:lens) =
...
              dels ">" . opt_eol . ((body|comment) . (body|empty|comment)*)? .
              indent . dels "</" in
-    let kword = key word in
-    let dword = del word "a" in
+    let kword = key (word - /perl/i) in
+    let dword = del (word - /perl/i) "a" in
         [ indent . dels "<" . square kword inner dword . del />[ \t\n\r]*/ ">\n" ]
 
+let perl_section = [ indent . label "Perl" . del /<perl>/i "<Perl>"
+                   . store /[^<]*/
```
raphink: Note: this is probably not a good idea… we'll probably get a new issue on that soonish…

raphink:
Merged as 3a901a4ba1166703bb0fc87d8dd9b7357a06d427

PR #338, opened 2015-12-21
Yaml subset lens

kurkale6ka:


lenses/yaml.aug
```diff
@@ -0,0 +1,53 @@
...
+  repo2: branch
+*)
+let entry = [ key Rx.word . colon . (space . anchor)? . eol
+            . (indent . comment)*
+            . ((inherit . (repo+)?) | repo+)
+            ]
+
+let lns = (entry | comment | empty)+
+
+let filter = incl "/path/to/custom.yml" . Util.stdexcl
```
raphink: `filter` and `xfm` are not required, and there's no point in have them here. Also, you can get rid of `autoload`.

lenses/yaml.aug
```diff
@@ -0,0 +1,53 @@
...
+
+(* &anchor *)
+let anchor = Util.del_str "&" . val
+
+(*
+host:
+  # Inheritance
+  <<: *anchor
+  repo2: branch
+*)
```
raphink: You could document the lens using NaturalDocs? See keepalived.aug for example.

raphink:
Merged as 2aeb42f7a5b0c7e2de725d5a640149de3be52112

PR #306, opened 2015-12-22
Invalid rules to parse /etc/network/interfaces

raphink:
Current rule contains unconditional parse of interfaces.d subdirectory. But it is not valid, because Debian parses it only when instruction 'source-directory interfaces.d' specified in the /etc/network/interfaces file. Moreover, parser does not understand 'source-directory' instruction which causes empty tree.
Current rule: 
  let filter = (incl "/etc/network/interfaces") . (incl "/etc/network/interfaces.d/*") . Util.stdexcl


raphink:
Attached code to fix.

Issue #367, opened 2016-04-22
logrotate lens does not support dateyesterday


chris-reeves:
Logrotate supports the 'dateyesterday' option to use yesterday's instead of today's date to create the dateext extension.

The logrotate lens does not support setting this option.

raphink:
Did you mean to create a PR for your commit?

PR #368, opened 2016-04-22
logrotate: support dateyesterday option (GH #367)

chris-reeves:
Adds support for the 'dateyesterday' directive for logrotate.

I didn't add any tests for this as it is a trivial addition, and there are no existing tests for the other flags. I have, however, tested this locally with no issues. I'm happy to submit a new pull request with tests if that is preferred.


raphink:
Lgtm, I'll merge this next week.

PR #378, opened 2016-06-09
Recent ntp.conf allows for 'pool' instead of 'server' et al

craigmiskell:
Seen on Ubuntu Xenial (16.04), and not in the man page, but found references at https://www.freebsd.org/cgi/man.cgi?query=ntp.conf&sektion=5.


raphink:
Thank you.

PR #379, opened 2016-06-09
Support for rsyslog RainerScript syntax

craigmiskell:
Specifically configuration objects per http://www.rsyslog.com/doc/v8-stable/rainerscript/configuration_objects.html

This is my first moderate-scale attempt at lense updating (not just adding keywords to existing regexes), so I welcome any advice/feedback on doing it better.


raphink:
Thank you. Merged as e00c7a815cb6c3e63aa3eec7329126fd2eccbf79 with a few modifications (style related).

PR #380, opened 2016-06-16
Add lens to parse postfix password maps

spacedog:
Postfix SMTP client can be configured to send username and password information to the mail gateway server. 

This lens parses postfix password map tables. More details [here](http://www.postfix.org/postconf.5.html#smtp_sasl_password_maps)


raphink:
LGTM

PR #386, opened 2016-07-11
fix restrict to allow also -4 and also fix save/store ability

jreidinger:
Restrict now can also have -4, so ipv6 no longer have sense alone, so changed to "ip" key with value "4" or "6". Also tests are adapted to new behavior

### DIFF
--- a/lenses/ntp.aug
+++ b/lenses/ntp.aug
@@ -67,9 +67,10 @@ module Ntp =
 
     (* Define restrict *)
     let restrict_record   =
-      let ip_restrict = [ label "ip" . sep_spc . Util.del_str "-" . store /[46]/ ] in
+      let ip6_restrict = [ label "ipv6" . sep_spc . Util.del_str "-6" ] in
+      let ip4_restrict = [ label "ipv4" . sep_spc . Util.del_str "-4" ] in
       let action    = [ label "action" . sep_spc . store /[^,# \n\t-][^,# \n\t]*/ ] in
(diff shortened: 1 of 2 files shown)


raphink:
OK.

Just a few thoughts:

* Could you add a test for "ipv4", too?
* This is a breaking change. Another option would be to keep the `ipv6` node and add an `ipv4` node option. @lutter, @domcleal any thoughts on that?

jreidinger:
@raphink yep, I will add it

domcleal:
> This is a breaking change. Another option would be to keep the ipv6 node and add an ipv4 node option

That sounds like a better idea to me, treat it as a flag rather than node with a `4` or `6` value.

jreidinger:
test added
btw there is also one another missing option. Restrict can have also optional mask after address ( but only if default is not used ), so maybe it make more sense to have it all as nodes like
```
restrict => {
  "address" => "5.5.5.5",
  "mask" => "255.255.0.0",
  "ip" => "4",
  "action[1]" => ...
}
```
but we do not use mask too much, so just side note.

jreidinger:
> That sounds like a better idea to me, treat it as a flag rather than node with a 4 or 6 value.

if there is agreement on it, I can change it.

raphink:
Yes, I'd rather go with the other option, for the sake of compatibility.

jreidinger:
updated to be backward compatible as agreed.

jreidinger:
thanks

raphink:
no, thank you :smile:

PR #388, opened 2016-07-22
DEREF_OF_NULL

g-coder:
Return value of a function 'add_state' is dereferenced at lens.c:1558 without checking.
Added check on return value.

### DIFF
--- a/src/lens.c
+++ b/src/lens.c
@@ -1556,6 +1556,8 @@ static struct prod *make_prod(struct rtn *rtn, struct lens *l) {
     ERR_BAIL(l->info);
 
     result->end->next = rtn->states;
+    if (result->end == NULL)
+        goto error;
     rtn->states = result->start;
 
     return result;


src/lens.c
```diff
@@ -1556,6 +1556,8 @@ static struct prod *make_prod(struct rtn *rtn, struct lens *l) {
     ERR_BAIL(l->info);
 
     result->end->next = rtn->states;
+    if (result->end == NULL)
+        goto error;
```
lutter: This check is not needed - though it's not entirely obvious: `add_state` calls `ERR_NOMEM` when allocation fails, which sets an error code in the `struct error` of `prod->lens->info`. `make_prod` uses `ERR_BAIL` to return early if any error has been encountered - the tricky bit is that there really is only on `struct error` in the system, so that `l->info->error` points to the same `struct error` that `prod->lens->info` points to.

lutter:
This is not needed.

PR #389, opened 2016-07-22
DEREF_OF_NULL

g-coder:
Return value of a function 'add_state' is dereferenced at fa.c:2776 without checking.
Added check on return value.


lutter:
Nice catch ! Thanks for the patch.

PR #390, opened 2016-07-22
NO_VA_END

g-coder:
Function 'va_end' was not called before internal.c:55 and internal.c:64 inside function 'pathjoin'.
Added 'va_end(ap)' before return statement.


lutter:
Thanks for the patch ! Merged as 057e9b3c

PR #391, opened 2016-07-24
Uses spaces between tokens instead of tabs for pam files

### LINKED ISSUE #236: pam lens: arguments should be separated by spaces not tabs?
augeasproviders_pam has made a horrible mess of my Oracle Linux files it touched.  Indentation is all over the place.

I think the problem is with the lens, but am not sure what the 'fix' might be.
Currently, the 'service', 'type', 'control', 'module-path' tokens *and* the arguments are all separated by tabs when a line is written by augeas.

I've compared various Ubuntu and OracleLinux pam.d files.
For the aforementioned tokens, sometimes they are tab separated and sometimes space separated.
However, the arguments always seem to be space separated. Should the lens be changed to mimic this?

What about the other whitespace?  Is there any way I can make this consistent across all entries in a pam.d file touched by augeas (with puppet)?

Many thanks,
Alex

lorengordon:
pam.d(5) man page states: "The format of each rule is a space
separated collection of tokens..." Previsouly, pam.aug used
`Util.del_ws_tab` to insert tabs as a token separator. This patch
modifies pamconf.aug and pam.aug to use `Sep.space`, which
uses a space as a separator instead.

Fixes #236

### DIFF
--- a/lenses/pam.aug
+++ b/lenses/pam.aug
@@ -48,11 +48,11 @@ module Pam =
   (* Shared with PamConf *)
   let record = [ label "optional" . del "-" "-" ]? .
                [ label "type" . store types ] .
-               Util.del_ws_tab .
+               Sep.space .
                [ label "control" . store control] .
-               Util.del_ws_tab .
+               Sep.space .
                [ label "module" . store word ] .
-               [ Util.del_ws_tab . label "argument" . store argument ]* .
+               [ Sep.space . label "argument" . store argument ]* .
                comment_or_eol
 
   let record_svc = [ seq "record" . indent . record ]
--- a/lenses/pamconf.aug
+++ b/lenses/pamconf.aug
@@ -39,7 +39,7 @@ let service = Rx.word
 
 let record  = [ seq "record" . indent .
               [ label "service" . store service ] .
-              Util.del_ws_tab .
+              Sep.space .
               Pam.record ]
 
 let lns = ( empty | comment | include | record ) *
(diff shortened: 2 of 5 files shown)


lorengordon:
I think I'd generally prefer to either set a default separator token and override it via a parameter, or to align the tokens to columns. If anyone can point me in the right direction for getting there, instead of just using a single space as a separator, I'm happy to modify this PR.

lorengordon:
Is this test failure valid? Attempting to get more information, I built my branch locally and ran `make check` but I'm not sure how to trace the output to my changes...

```
make[2]: Entering directory `/root/augeas/tests'
  CC     fatest.o
cc1: warnings being treated as errors
fatest.c: In function ‘testEnumerate’:
fatest.c:654: error: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result
make[2]: *** [fatest.o] Error 1
make[2]: Leaving directory `/root/augeas/tests'
make[1]: *** [check-am] Error 2
make[1]: Leaving directory `/root/augeas/tests'
make: *** [check-recursive] Error 1
# echo $?
2
````

lorengordon:
Alright, I figure it must have something to do with the test fixtures under `tests/test-augtool`, as the files and diffs are written with tabs, but I can't for the life of me figure out which of the tests is actually failing or how to adjust them to get them to work. I'm a bit frustrated, so I'm going to step away for a while. Hopefully someone can chime in and show me how to figure this out.

lorengordon:
Ok, I think that error was a result of following the build directions in HACKING.md, using `./autogen.sh --enable-compile-warnings=error --enable-debug=yes`. When I follow the setup in .travis.yml and just use `./autogen.sh` without any arguments, I get past that error. However, then `make check` hangs just after passing the `lens-kdump.sh` test. So, now I'm resorting to manually running the test fixtures using `./src/try`, inspecting the result, updating the test diff, crossing my fingers, and pushing to let travis test it.

lorengordon:
Oh, and hurray! Tests just passed!

raphink:
@lutter are you fine with that?

lutter:
I am totally fine with changing the default if that causes fewer problems. Merged as 3a1c7fd3

PR #392, opened 2016-07-25
Shellvars: include lbu.conf

kunkku:
Alpine Linux Local Backup Utility


lutter:
Thanks for the patch. Merged as 4be260f5

PR #401, opened 2016-09-05
Add Ceph module

Paulche:
A simple bunch of tests is added as well.

### DIFF
--- a/lenses/ceph.aug
+++ b/lenses/ceph.aug
@@ -1,0 +1,26 @@
+(* Ceph module for Augeas
+ Author: Pavel Chechetin  <pchechetin@mirantis.com>
+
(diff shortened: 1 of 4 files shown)


raphink:
That looks good. Could you just add NaturalDocs documentation to the lens please? (see `keepalived.aug` for example).

Paulche:
Good point! I'll add the documentation soon. Thanks!

pspacek:
@Paulche Could you add the documented lens to this PR? Thanks!

lutter:
@raphink any objections to merging this as-is ? I'd rather have this in without docs if it works

lutter:
Merged as 1a3b4889

Issue #397, opened 2016-08-16
Double free with AUG_ENABLE_SPAN in free_tree_node


theopolis:
I tried a tiny test:
```cpp
#include <augeas.h>

int main() {
  augeas *aug = aug_init(nullptr, nullptr, 0);

  if (aug == nullptr) {
    return 1;
  } else {
    aug_close(aug);
  }

  return 0;
}
```

```
g++ -I /usr/local/include -std=c++11 -L /usr/local/lib ./main.cpp -l augeas
./a.out
```

Cool!

Now change:
```cpp
  augeas *aug = aug_init(nullptr, nullptr, AUG_ENABLE_SPAN);
```

And it will crash with the double free. Perhaps I'm using the C API incorrectly?

```
*** Error in `a.out': double free or corruption (fasttop): 0x0000000006353310 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x77725)[0x7ff0d1910725]
/lib/x86_64-linux-gnu/libc.so.6(+0x7ff4a)[0x7ff0d1918f4a]
/lib/x86_64-linux-gnu/libc.so.6(cfree+0x4c)[0x7ff0d191cabc]
a.out[0x16657c1]
a.out(free_tree+0x28)[0x1666d30]
a.out(free_tree+0x1b)[0x1666d23]
a.out(free_tree+0x1b)[0x1666d23]
a.out(free_tree+0x1b)[0x1666d23]
a.out(free_tree+0x1b)[0x1666d23]
a.out(free_tree+0x1b)[0x1666d23]
a.out(aug_close+0x15)[0x1668656]
```

lutter:
I just tried this on my machine (Fedora 23, augeas 1.5.0) and it does not produce an error. Running it under valgrind or with `MALLOC_CHECK_=2` in the environment doesn't turn up anything, either. Since you are pulling in stuff from `/usr/local`, what do you have in there ? Just augeas (and what version) ?

It could be that some config file you have triggers something. Can you try if running `AUGEAS_ROOT=/dev/null ./a.out` triggers the error, too ? Also, since it's roughly equivalent, what happens when you run `augtool --span quit` ?

theopolis:
Cool! I'll try on a fresh VM with debugging enabled for the few dependencies and see if I can reproduce.

theopolis:
[...]

PR #403, opened 2016-09-26
Fix memory corruption when using spans with a recursive lens (bug #397)

lutter:
The problem described in bug #397 is caused by an error handling memory when spans are enabled. The fix for this problem is the last patch in hte series, the others were merely minor unrelated things I encountered when fixing this bug.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,8 @@
+1.7.0 - ???
+  - General changes/additions
+    * fix a use-after-free in recursive lenses when spans are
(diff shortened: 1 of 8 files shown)


thedrow:
Can you please merge this? We want to release the next osquery version with the augeas table included. We have our own build system and we can use a specific commit.
ping @lutter @raphink

lutter:
@thedrow Just merged this. Please confirm that it works for you once you have tested it.

PR #404, opened 2016-09-29
Run with address sanitizer in CI

thedrow:
This will prevent bugs like #403 from ever hitting master.
The address sanitizer is able to detect use-after-free issues and other bad memory handling issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,17 @@
-sudo: false
-addons:
-  apt:
-    packages:
-      - libxml2-dev
-      - libreadline-dev
-      - valgrind
+sudo: true
+dist: trusty
+cache:
+  ccache: true
+  directories:
+    - $TRAVIS_BUILD_DIR/gnulib
 install:
-  - ./autogen.sh
-  - sed -i '41s/ENOENT/ENOENT || errno == EINVAL/' gnulib/tests/test-readlink.h
+  - ./ci/install.sh
 language: c
+compiler:
+  - gcc-6
 notifications:
   email: false
   irc:
     channels:
       - "irc.freenode.org#augeas"
-script: ./configure && make && make check && ./src/try valgrind
+script: ./ci/test.sh
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -35,7 +35,7 @@ sudo pip3 install -q colout2
 end_fold_marker 'dependencies'
 
 start_fold_marker 'autogen'
-./autogen.sh 2>&1 | colout2 -t autogen
+./autogen.sh --enable-asan 2>&1 | colout2 -t autogen
 exit_status=$?
 end_fold_marker 'autogen'
 
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -o pipefail
 
 start_fold_marker() {
   echo "travis_fold:start:install.$1"
--- a/ci/install.sh
+++ b/ci/install.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+start_fold_marker() {
+  echo "travis_fold:start:install.$1"
+}
+
+end_fold_marker() {
+  echo "travis_fold:end:install.$1"
+}
+
+start_fold_marker 'dependencies'
+if [[ $CC = "gcc"* ]];
+then
+  sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
+# elif [[ $CC = 'clang'* ]]; then
+#   sudo apt-add-repository 'deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-${CC#clang-} main'
+#   wget -O - http://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+fi
+sudo add-apt-repository ppa:jonathonf/binutils -y &&\
+  sudo add-apt-repository ppa:ondrej/autotools -y
+
+sudo apt-get -qq update
+
+sudo apt-get install aria2
+git clone https://github.com/ilikenwf/apt-fast /tmp/apt-fast --depth 1
+sudo cp /tmp/apt-fast/apt-fast /usr/bin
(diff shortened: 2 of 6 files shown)


.travis.yml
```diff
@@ -1,17 +1,13 @@
-sudo: false
-addons:
-  apt:
-    packages:
-      - libxml2-dev
-      - libreadline-dev
-      - valgrind
+sudo: true
+dist: trusty
```
lutter: I am not too familiar with Ubuntu's releases, but why base this on Trusty rather than Xenial ?

ci/install.sh
```diff
@@ -0,0 +1,35 @@
...
+}
+
+end_fold_marker() {
+  echo "travis_fold:end:install.$1"
+}
+
+start_fold_marker 'dependencies'
+sudo add-apt-repository ppa:jonathonf/binutils -y &&\
+  sudo add-apt-repository ppa:ondrej/autotools -y &&\
+  sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
```
lutter: Again, my lack of familiarity with Ubuntu, but I am not really comfortable with using autotools that are not in the base distro.

configure.ac
```diff
@@ -61,6 +61,7 @@ AC_ARG_ENABLE([debug],
                              [enable debugging output])],[],[enable_debug=yes])
 AM_CONDITIONAL([ENABLE_DEBUG], test x"$enable_debug" = x"yes")
 if test x"$enable_debug" = x"yes"; then
+   CFLAGS="${CFLAGS} -O0 -g3 -fsanitize=address -fno-omit-frame-pointer"
```
lutter: Running `make check` with address sanitizer on makes running tests on my machine _a lot_ slower. We should put turning ASAN on behind a different flag, say `--enable-asan`, or just pass `CFLAGS` and `LDFLAGS` manually from `test.sh` (which I would call `setup.sh` or some such, since it's not really a test, but setup for tests)

lutter:
@thedrow Have a look at PR #405 which have lots of fixes - it doesn't make a build with ASAN succeed yet, but seems to fix a few problems.

thedrow:
@lutter I finally managed to run with ASAN. This is mergeable.

.travis.yml
thedrow: Travis only supports precise or trusty.

ci/install.sh
thedrow: It's fine for CI in my opinion. I can remove it though.

configure.ac
thedrow: I think it's mainly because I specified -O0. I can increase the optimization level.

thedrow:
@lutter Let me know when you merge #405 so I can rebase.

thedrow:
I'm still working on this. Don't merge this yet.

configure.ac
lutter: It's not just because of `-O0` - it really needs to either go behind a separate configure flag, or just run `configure CFLAGS='..' LDFLAGS='..'` from `test.sh`

lutter:
It'll probably be a while before we can merge this - I've been looking at the various failures that ASAN produces, and there's a couple leaks that will be pretty fun to fix (none of them are big, i.e. it'll take quite some time to leak 1kB, but we need to fix them nonetheless)

Could you also squash this PR into one commit or a small number of commits ?

thedrow:
I will squash my PR as soon as I finish working on it :)

thedrow:
@lutter I disabled testing with Clang for now. I'll address it in another PR. Since this passes do you want to merge this or do you want me to adjust something?

thedrow:
I am really struggling to figure out what's wrong with Ubuntu's environment in such a way that it won't find libasan or pthread.

configure.ac
```diff
@@ -61,6 +61,8 @@ AC_ARG_ENABLE([debug],
                              [enable debugging output])],[],[enable_debug=yes])
 AM_CONDITIONAL([ENABLE_DEBUG], test x"$enable_debug" = x"yes")
 if test x"$enable_debug" = x"yes"; then
+   CFLAGS="${CFLAGS} -O0 -g3 -fsanitize=address -fno-omit-frame-pointer -lasan -lpthread"
+   LDFLAGS="${LDFLAGS} -fsanitize=address -lasan"
```
lutter: Can we please not make `--enable-debug` also turn on ASAN ? Other options would be to either run `./configure` in the setup script passing the appropriate `CFLAGS` and `LDFLAGS` or putting it behind a `--enable-asan` flag or having a `make asan` target that builds with those flags - all of these are fine by me, but I don't want it to be the default when you run with debugging on.

lutter:
This is really weird. I have ASAN working locally through running `./configure  CFLAGS='-fsanitize=address -g' LDFLAGS=-fsanitize=address` and `make check` succeeds for me. We definitely need a successful build before we can merge this.

configure.ac
thedrow: I'll do so promptly.

thedrow:
It seems like valgrind doesn't play nice with ASAN for some reason. I'm going to try to separate them.

ci/test.sh
```diff
@@ -0,0 +1,64 @@
...
+set -e
+start_fold_marker 'test.lenses'
+start_fold_marker 'test.lenses.asan'
+./src/try asan
+end_fold_marker 'test.lenses.asan'
+start_fold_marker 'test.lenses.valgrind'
+make clean
+./autogen.sh --enable-debug 2>&1 | colout2 -t autogen
+make -j 2
+./src/try valgrind 2>&1 | colout2 -t valgrind
```
lutter: I actually think that if we run with ASAN, we can skip running with valgrind. The reason we did this is to get some simple info about leaks etc. and I think we get much better coverage of the same thing with ASAN now.

thedrow: I'm not sure but I think valgrind is able to detect leaks that asan isn't able.

thedrow:
I'm going to close this since I don't know how to make this work.
We'll try again some day.

PR #408, opened 2016-10-03
Support for pgpool.conf in postgresql lens

loopiajoakim:
This lens mostly works for pgpool.conf too but fails on integers and single words. Replaced Quote.do_squote_opt with Quote.do_squote_opt_nil in lenses/postgresql.aug and now it seems to work.

### DIFF
--- a/lenses/postgresql.aug
+++ b/lenses/postgresql.aug
@@ -35,7 +35,7 @@ let word_opt_quot_re = /[A-Za-z][A-Za-z0-9_]*/
 
 (* View: word_opt_quot
      Storing a <word_opt_quot_re>, with or without quotes *)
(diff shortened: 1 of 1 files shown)


lutter:
Thanks for the contribution - the build fails because one of the tests in `lenses/tests/test_postgresql.aug` fails. Can you fix that failure ?

raphink:
I'm not sure if the compatibility with postgresql.conf is guaranteed. I'd be more in favour of a pgpool lens based on the postgresql one.

loopiajoakim:
I kind of agree that a new lens might be better. I would say I'm not the one to write it as I'm not that good at Augeas yet.

raphink:
You can start with a pgpool module that uses `Postgresql.lns`

loopiajoakim:
I will try it when there is time, thanks :)

PR #410, opened 2016-10-04
Refactorings/stability fixes for recursive parsing

lutter:
The code that dealt with parsing files with recursive lenses had a number of issues that are fixed by this PR. In particular, the handling of the parse stack was too complicated, and there were a few places where an error might lead to leaking memory.

### DIFF
--- a/src/errcode.h
+++ b/src/errcode.h
@@ -91,24 +91,19 @@ void reset_error(struct error *err);
         }                                                       \
     } while(0)
 
-/* A variant of assert that uses our error reporting infrastructure
- * instead of aborting
+/* Assertions that use our error reporting infrastructure instead of
+ * aborting
  */
-#ifdef NDEBUG
-# define ensure(cond, obj) if (0) goto error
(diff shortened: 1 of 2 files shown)


src/errcode.h
```diff
@@ -91,24 +91,19 @@ void reset_error(struct error *err);
...
-#else
-# define ensure(cond, obj)                                           \
+#define ensure(cond, obj)                                            \
     if (!(cond)) {                                                   \
         bug_on((obj)->error, __FILE__, __LINE__, NULL);              \
         goto error;                                                  \
     }
-# define ensure0(cond, obj)                                          \
+#define ensure0(cond, obj)                                           \
     if (!(cond)) {                                                   \
```
thedrow: We should probably use `builtin_expect` if it's available in our platform.

lutter: I don't think the use of `ensure0` is a big deal performance-wise. We can add some optimizations if that seems warranted, but for now, I'll commit this as is

PR #412, opened 2016-10-09
First draft of opendkim config file support

craigmiskell:
Supports /etc/opendkim.conf (known good on OpenDkim 2.9.1 as per Ubuntu 14.04.  

This is my first from-scratch lens, so there may be style problems etc; I welcome any suggestions or other feedback.

### DIFF
--- a/lenses/opendkim.aug
+++ b/lenses/opendkim.aug
@@ -1,0 +1,25 @@
+module Opendkim =
+  autoload xfm
+
+  (* Inifile.comment is saner than Util.comment regarding spacing after the # *)
+  let comment  = Inifile.comment "#" "#" 
+  let eol = Util.eol
+  let empty = Util.empty
+
+  (*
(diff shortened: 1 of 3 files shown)


lutter:
This looks great - the main reason why I haven't merged that is that the subject says 'First draft'; there's only a few cosmetic changes that I would want to make:

* Squash the two commits into one (yes, you need the second commit)
* Change the subject of the commit message to the format we use for lenses, "Opendkim: new lens for ..."
* Reformat opendkim.aug to have lines with 80 characters max

For the last point it might help that you can write stuff like
```
let stringv = key ( /a|b|c/ .
                           /d|e|f/) . del ...
```

In any event, thanks a ton for writing this lens ! I'd love to merge this after these minor changes.

craigmiskell:
Thanks; will open a fresh PR in one commit, with style changes requested (after squashing I'm not at all clear how to get that change onto this PR)

PR #418, opened 2016-10-12
Opendkim: new lens for /etc/opendkim.conf

craigmiskell:
New lens for Opendkim, specifically /etc/opendkim.conf

### DIFF
--- a/lenses/opendkim.aug
+++ b/lenses/opendkim.aug
@@ -1,0 +1,57 @@
+module Opendkim =
+  autoload xfm
+
+  (* Inifile.comment is saner than Util.comment regarding spacing after the # *)
+  let comment  = Inifile.comment "#" "#" 
+  let eol = Util.eol
+  let empty = Util.empty
+
+  (*
+    The Dataset spec is so broad as to encompass any string (particularly the
+    degenerate 'single literal string' case of a comma separated list with 
+    only one item).  So treat them as 'String' types, and it's up to the user to
+    format them correctly.  Given that many of the variants include file paths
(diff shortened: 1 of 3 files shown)


lutter:
thanks for the contribution ! Unfortunately, when I run this through augparse with `augparse lenses/opendkim.aug` I get the following error:
```
lenses/opendkim.aug:16.21-.75:Invalid regular expression: Unmatched ( or \(
lenses/opendkim.aug:35.22-.63:Invalid regular expression: Unmatched ( or \(
lenses/opendkim.aug:41.22-.64:Invalid regular expression: Unmatched ( or \(
lenses/opendkim.aug:52.16-.25:Undefined variable integerkv
lenses/opendkim.aug:54.31-.36:Undefined variable entry
lenses/opendkim.aug:56.22-.25:Undefined variable lns
```

The problem seems to be that you split regular expressions across lines like so: `key /(a|b/ . /c|d)/`. That doesn't work since everything between a pair of `/.../` needs to be a valid regular expression. Since the entire expression is just a union, the parentheses are actually unnecessary and you should write `/a|b/ | /c|d/` instead.

lutter:
I actually made the changes I suggested and commited this now as 8eb61f2f

Thanks again for the contribution !

craigmiskell:
Ahh, thank you.  I saw the failure, but couldn't figure it out and haven't had the time to debug in depth.  Thanks for fixing it.

lutter:
Would have been easier if some chump hadn't given you bad advice on the first PR ;)

PR #425, opened 2016-11-04
Refine handling of multiple transforms handling the same file

lutter:
We used to blindly flag an error when there were multiple transforms that
handled the same file. That is now ok as long as these transforms all use
the exact same lens.


rwmjones:
I'm afraid I don't have any very detailed tests, however I did apply this patch and rerun the tests described in https://www.redhat.com/archives/augeas-devel/2016-November/msg00004.html and those now work fine, so it looks good to me.

Issue #430, opened 2016-12-20
RHEL7.3 krb5-libs package comes with config containing includedir.


smithj4:
Last month RedHat released an update of RHEL7 to the .3 version. The krb5-libs package now comes with a default /etc/krb5.conf config file that begins with:

    # Configuration snippets may be placed in this directory as well
    includedir /etc/krb5.conf.d/

The Augeas krb5 lens does not understand the includedir config keyword and throws an error.

PR #431, opened 2016-12-21
Possible fix to #430

smithj4:
I don't know Augeas lenses that well, but looking at some others that have includes in them as examples, this might be a possible fix to Issue #430 that I submitted.


domcleal:
Can you also add a test case to lenses/tests/test_krb5.aug please?

domcleal:
Merged, thanks @smithj4.

PR #432, opened 2016-12-23
Removed trailing space in token in slapd.lens

iavael:


domcleal:
Merged, thanks @iavael!

PR #433, opened 2016-12-29
php.aug  - Fix php7 FPM support

ddico:
Update path to reflect the transition from:
/etc/php5/ to /etc/php/7.0/


domcleal:
Merged, thanks @ddico.

PR #436, opened 2017-02-01
Chrony: add new directives and options

mlichvar:
Add new directives and options from chrony-3.0 and chrony-3.1.


raphink:
Thank you

PR #437, opened 2017-02-04
[krb5.aug] Support realms that start with numbers

mdwheele:
Currently, the default kerberos configuration that ships with
Ubuntu has a realm that starts with a number (1TS.ORG). This
causes the parser to fail and prevents krb5.conf from being
available via augtool.

This patch allows numbers 0-9 as the first character of a
realm.

```
[realms]
~~<snip>~~
    1TS.ORG = {
        kdc = kerberos.1ts.org
        admin_server = kerberos.1ts.org
    }
```

### DIFF
--- a/lenses/krb5.aug
+++ b/lenses/krb5.aug
@@ -21,8 +21,8 @@ let closebr = del /[ \t]*\}/ "}"
    and realms in the [appdefaults] section.
 *)
 
-let realm_re = /[A-Z][.a-zA-Z0-9-]*/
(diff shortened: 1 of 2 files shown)


mdwheele:
Hmm, probably should have added my example to the test... duh.

mdwheele:
If you would like me to create a test specific for this rather than putting the failing case in what looked to be more of a end-to-end "does it load a full config", let me know and I can do that. If so, let me know what the best way of doing that is (otherwise, I'd be copy-pasta'ing to success) 😄

raphink:
Thank you.

PR #440, opened 2017-02-14
src/augtool: fix multiple definition of xasprintf

joerg-krause:
`xasprintf` is defined both in *internal.c* and *augtool.c*. This breaks building augeas statically:

```
./.libs/libaugeas.a(internal.o): In function `xasprintf':
internal.c:(.text+0x934): multiple definition of `xasprintf'
augtool.o:augtool.c:(.text+0x8a4): first defined here
collect2: error: ld returned 1 exit status
```

Make the definition of `xasprintf` in *augtool.c* a private copy by prefixing the function name with an underscore and declaring the function as static.

### DIFF
--- a/src/augtool.c
+++ b/src/augtool.c
@@ -65,8 +65,10 @@ char *history_file = NULL;
  * General utilities
  */
 
-/* Not static, since prototype is in internal.h */
-int xasprintf(char **strp, const char *format, ...) {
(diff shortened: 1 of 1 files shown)


lutter:
I actually stumbled over this a while ago, but didn't fix it. Thanks for fixing it ! Committed as 05a27f4e

PR #441, opened 2017-02-15
Updated for multipath-0.4.9-99.el7.

XMol:


domcleal:
You can reproduce the test failures locally either with `make check`, or by just running `augparse -I lenses lenses/tests/test_multipath.aug`. It's failing to parse something in the defaults group at around line 31 I would guess.

(You can also add new examples to this test file to cover any new functionality you've added.)

XMol:
Hi Dominic,

there is no test_multipath.aug lens - does that explain the failed check?

Ciao,
Xavier.

domcleal:
It's here, in the git tree: https://github.com/XMol/augeas/blob/b8af6f930b05a459bc4c353fc59fbfe28c6534c3/lenses/tests/test_multipath.aug

XMol:
Yes, I just found it, too. Was looking into the wrong tests directory...

XMol:
Ok, added some more undocumented attributes, that were included in the test lens.

Of course, whatever isn't mentioned in the man pages, but was implemented by the vendors for their individual products, is difficult to test in advance...

domcleal:
Merged, thanks for the patch @XMol.

PR #442, opened 2017-02-16
Update interfaces.aug

joerg-krause:
Fix typos.


domcleal:
Merged, thank you @joerg-krause.

Issue #439, opened 2017-02-14
Allow underscores in postfix_virtual


lingfish:
As far as I can tell, this is legal in an e-mail address; postfix accepts it fine.

Similiar to https://github.com/hercules-team/augeas/commit/c3c7bfa8f7b042be417b39889187c8b37b3297ee, all that is needed is:

```
--- /usr/share/augeas/lenses/dist/postfix_virtual.aug.ORIG      2016-01-05 10:02:36.000000000 +1100
+++ /usr/share/augeas/lenses/dist/postfix_virtual.aug   2017-02-14 12:37:34.024967401 +1100
@@ -32,7 +32,7 @@
   del (space_or_eol_re? . sep . space_or_eol_re?) default

 (* View: word *)
-let word = store /[A-Za-z0-9@\*.+=-]+/
+let word = store /[A-Za-z0-9@\*.+=-_]+/

 (* View: comma *)
 let comma = space_or_eol "," ", "

```

domcleal:
If you can open that as a pull request then we'd be happy to merge it. A bonus would be a small example also added to lenses/tests/test_postfix_virtual.aug.

lingfish:
You're right... I tried to sneak a lazy through ;)  Done.

PR #445, opened 2017-02-19
Allow underscores in postfix_virtual

lingfish:
As per #439, understores are legal in an e-mail address.

### DIFF
--- a/lenses/postfix_virtual.aug
+++ b/lenses/postfix_virtual.aug
@@ -32,7 +32,7 @@ let space_or_eol (sep:regexp) (default:string) =
   del (space_or_eol_re? . sep . space_or_eol_re?) default
 
 (* View: word *)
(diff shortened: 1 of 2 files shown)


lenses/postfix_virtual.aug
```diff
@@ -32,7 +32,7 @@ let space_or_eol (sep:regexp) (default:string) =
   del (space_or_eol_re? . sep . space_or_eol_re?) default
 
 (* View: word *)
-let word = store /[A-Za-z0-9@\*.+=-]+/
+let word = store /[A-Za-z0-9@\*.+=-_]+/
```
lutter: You need to keep the dash `-` as the last thing in that character class. The way you have it now, this gets interpreted as 'include all characters from `=` to `_`' which is not what you want.

domcleal:
Merged, thanks @lingfish!

PR #446, opened 2017-02-20
Add lens for radicale config file

jvalleroy:
This lens can parse /etc/radicale/config. It's a standard INI file, so the lens is basically the same as Puppet lens, except for the included path.


raphink:
Thank you for this. Could you please add a unit test file?

jvalleroy:
@raphink I've added a unit test.

lutter:
Looks good to me. Thanks for the contribution

Issue #384, opened 2016-07-03
Add an API to get the filename without parsing spans


thedrow:
Is it possible to get the filename without parsing spans?
If it is, can we expose an API that does that?

raphink:
You can get that info without parsing spans if you look in `/augeas/files`, find a matching `path` node by removing `/files` in front.

thedrow:
Let's document it then 😄

thedrow:
While useful that doesn't fit my usecase (unless I'm doing something wrong).
I need to get the filename from any path expression that returns a single item.

```
match /augeas/files//*[path = "/files/etc/hosts/1/ipaddr"]
(no matches)
```

but
```
match /augeas/files//*[path = "/files/etc/hosts"]
/augeas/files/etc/hosts = (none)
```

raphink:
That's right, it's not very simple indeed.

thedrow:
I see that the info struct has a filename char*. Can we use it if we're not enabling spans?
The problem is that parsing spans takes more CPU which isn't always needed.

lutter:
The problem with spans is that they make the entire tree take quite a bit more memory; that's the reason why they are disabled by default. How much additional CPU usage do you see ? This will most likely be due to additional allocation (the information that goes into spans is not very hard to collect internally)

If you really can't use spans for your purpose, and since querying the tree directly will be pretty hairy, we could add a call to the API, something like `aug_source_file(augeas *, const char *path, char **filename)`

How often do you need to query the filename for a node ?

thedrow:
I query it once per node (and then cache) since I want users to be able to filter by filenames as well.
I see roughly 10% more CPU usage when using spans and when parsing every single file we're already at the 60-90% range.

lutter:
I just looked at this again, and thought about how this would be implemented; it occurred to me that adding the XSLT function [starts-with](http://zvon.org/xxl/XSLTreference/Output/function_starts-with.html) would let you do what you want; i.e., you could then say

```
match /augeas/files//path[starts-with("/files/etc/hosts/1/ipaddr", self::path)]
```

If you need to do that for every node though, no matter what we do (including the `aug_source_file` I mentioned above) will not be terribly performant, and you might be better off building an index from `match /augeas/files//path` and checking nodes against that. That might in the end be the fastest way to achieve that, especially if you traverse the tree top-down, and use that fact that once you found the root for a file, all its children will also be from that file.

thedrow:
[...]

PR #448, opened 2017-02-24
Add a new aug_source API call and 'source' command in augtool

lutter:
Fixes https://github.com/hercules-team/augeas/issues/384


thedrow:
This is exactly what I need! Thanks.

PR #449, opened 2017-03-07
Reduce the total amount of memory used by Augeas

lutter:
The pattern buffers for compiled regular expressions take up a large amount
of space. Previously, we had placed calls to lens_release strategically so
that that memory gets released when we are done with a lens for a while
(for example, after loading all files described by a transform)

That overlooked that the interpreter compiles regular expressions when it
loads lenses. We now also release lens storage after loading an entire
module. That requires that regular expressions are compiled again when we
start actually loading files, but the additional regex compilation does not
seem to affect the overall runtime. It does dramatically reduce maximum
memory needed by an 'augtool -L quit'. On my machine, it drops from about
90MB to about 20MB.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,8 +1,16 @@
 1.8.0 - ???
   - General changes/additions
     * augtool: add a 'source' command exposing the aug_source API call
+    * augtool: dramatically reduce memory consumption when all lenses are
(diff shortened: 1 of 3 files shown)


raphink:
Wow looks pretty nice :-)

PR #452, opened 2017-03-15
Add 'context' and 'info' commands to augtool; reorganize help a little

lutter:


src/augrun.c
```diff
@@ -1116,6 +1148,49 @@ static const struct command_def cmd_load_def = {
...
+        if (entry != cmd->aug->modpathz) {
+            fprintf(cmd->out, ":");
+        }
+        fprintf(cmd->out, "%s", entry);
+    }
+    fprintf(cmd->out, "\n");
+
+    {
+
+    }
```
domcleal: Unused block?

lutter: Urgh .. yes, I wanted to print more stuff but ran out of steam

lutter:
I am not entirely sure how useful the `info` command is in its current incarnation .. maybe leave that out for now ?

domcleal:
It's not very useful, but if new functionality is added then it's a reasonable base to extend. It could include the path context, version of libfa etc?

lutter:
I added context and the number of files currently loaded .. unless you can think of anything else to throw in there, I'll merge that tomorrow.

PR #453, opened 2017-03-16
Fix for rsyslog rainerscript module loading, with options

craigmiskell:
Turns out module loads can have more than one option.  This fixes handling of that, and adds some related tests to be really sure.


lutter:
Thanksa ton for the contribution !

PR #463, opened 2017-05-09
Add new nslcd.conf lenses to parse nss-pam-ldapd configuration file

jplana:
Here's support for the /etc/nslcd.conf, the configuration file for nss-pam-ldapd (https://arthurdejong.org/nss-pam-ldapd/).

In the PR I've tried to include tests, an example file and some documentation that might help others use the lenses, but please let me know if something is missing.

### DIFF
--- a/lenses/tests/test_nslcd.aug
+++ b/lenses/tests/test_nslcd.aug
@@ -1,0 +1,408 @@
+(*
+Module: Test_Nslcd
+  Provides unit tests and examples for the <Nslcd> lens.
+*)
+
+module Test_nslcd =
+
+let real_file = "# /etc/nslcd.conf
+# nslcd configuration file. See nslcd.conf(5)
+# for details.
+
+# Specifies the number of threads to start that can handle requests and perform LDAP queries.
+threads 5
+
+# The user and group nslcd should run as.
+uid nslcd
+gid nslcd
+
+# This option controls the way logging is done.
+log syslog info
+
+# The location at which the LDAP server(s) should be reachable.
+uri ldaps://XXX.XXX.XXX
+
+# The search base that will be used for all queries.
+base dc=XXX,dc=XXX
+
+# The LDAP protocol version to use.
+ldap_version 3
+
+# The DN to bind with for normal lookups.
+binddn cn=annonymous,dc=example,dc=net
+bindpw secret
+
+
+# The DN used for password modifications by root.
+rootpwmoddn cn=admin,dc=example,dc=com
+
+# The password used for password modifications by root.
+rootpwmodpw XXXXXX
+
+
+# SASL authentication options
+sasl_mech OTP
+sasl_realm realm
+sasl_authcid authcid
+sasl_authzid dn:cn=annonymous,dc=example,dc=net
+sasl_secprops noanonymous,noplain,minssf=0,maxssf=2,maxbufsize=65535
+sasl_canonicalize yes
+
+# Kerberos authentication options
+krb5_ccname ccname
+
+# Search/mapping options
+
+# Specifies the base distinguished name (DN) to use as search base.
+base dc=people,dc=example,dc=com
+base dc=morepeople,dc=example,dc=com
+base alias dc=aliases,dc=example,dc=com
+base alias dc=morealiases,dc=example,dc=com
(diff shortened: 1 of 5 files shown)


lenses/tests/test_nslcd.aug
```diff
@@ -0,0 +1,408 @@
...
+  {  }
+  { "#comment" = "The FILTER is an LDAP search filter to use for a specific map." }
+  { "filter"
+    { "group" = "(objectClass=posixGroup)" }
+  }
+  {  }
+  { "#comment" = "This option allows for custom attributes to be looked up instead of the default RFC 2307 attributes." }
+  { "map"
+    { "passwd"
+      { "homeDirectory" = "\"${homeDirectory:-/home/$uid}\"" }
```
domcleal: Should quotes be removed here if present? The `Quote` module has lots of helpers for this, you can use things like `Quote.any_opt` to support store a value and optionally remove quotes if there are either double or single quotes.

lenses/tests/test_nslcd.aug
```diff
@@ -0,0 +1,408 @@
...
+  { "nss_initgroups_ignoreusers" = "user1,user2,user3" }
+  { "nss_min_uid" = "1000" }
+  { "nss_nested_groups" = "yes" }
+  { "nss_getgrent_skipmembers" = "yes" }
+  { "nss_disable_enumeration" = "yes" }
+  { "validnames" = "/^[a-z0-9._@$()]([a-z0-9._@$() \~-]*[a-z0-9._@$()~-])?$/i" }
+  { "ignorecase" = "yes" }
+  { "pam_authc_ppolicy" = "yes" }
+  { "pam_authz_search" = "(&(objectClass=posixAccount)(uid=$username)(|(authorizedService=$service)(!(authorizedService=*))))" }
+  { "pam_password_prohibit_message" = "\"MESSAGE LONG AND WITH SPACES\"" }
```
domcleal: ditto, looks like quotes should perhaps be removed automatically

lenses/tests/test_nslcd.aug
```diff
@@ -0,0 +1,408 @@
...
+  { "#comment" = "Specifies the ciphers to use for TLS." }
+  { "tls_ciphers" = "TLSv1" }
+  { "#comment" = "Specifies the path to the file containing the local certificate for client TLS authentication." }
+  { "tls_cert" = "/etc/ssl/certs/cert.pem" }
+  { "#comment" = "Specifies the path to the file containing the private key for client TLS authentication." }
+  { "tls_key" = "/etc/ssl/private/cert.pem" }
+  {  }
+  { "#comment" = "Other options" }
+  { "pagesize" = "100" }
+  { "nss_initgroups_ignoreusers" = "user1,user2,user3" }
```
domcleal: Should this be split into a list?

lenses/tests/test_nslcd.aug
jplana: I guess we can't remove the quotes for this specific field, as they are part of the value itself:

(from the documentation @ https://arthurdejong.org/nss-pam-ldapd/nslcd.conf.5)

> map MAP ATTRIBUTE NEWATTRIBUTE
> This option allows for custom attributes to be looked up instead of the default RFC 2307 attributes. The MAP may be one of the supported maps below. The ATTRIBUTE is the one as used in RFC 2307 (e.g. userPassword, ipProtocolNumber, macAddress, etc.). The NEWATTRIBUTE may be any attribute as it is available in the directory.
> If the NEWATTRIBUTE is presented in quotes (") it is treated as an expression which will be evaluated to build up the actual value used. See the section on attribute mapping expressions below for more details.

lenses/tests/test_nslcd.aug
jplana: Nice, I changed the tree be an indexed list.

lenses/tests/test_nslcd.aug
jplana: I did remove the quotes for this field, as they should be always present (in the configuration file)

jplana:
Reviewing the documentation I realized that the URL field could also be split into a list (separated by spaces instead of commas). I also fixed the documentation menu issue.

domcleal:
Merged, thanks @jplana!

PR #474, opened 2017-06-17
Add aug_get_nodes() to the API

thedrow:
Currently, the API does not allow fetching all the data for nodes matching a pattern. This addition to the API will allow fetching the path, label, value and possibly the filename of each node in a single tree traversal.
Right now in OSQuery we use `aug_get()` and `aug_label()` on each match we get from `aug_match()` which is inefficient and unnecessary.
`aug_get_nodes()` will return an array of structs with all the relevant information.

Feel free to come up with a better name.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -879,6 +879,25 @@ int aug_get_nodes(const struct augeas *aug, const char *pathin, struct aug_node 
         (*nodes)[i]->path = node_path;
         (*nodes)[i]->label = tree->label;
         (*nodes)[i]->value = tree->value;
+
+        struct span *span = tree->span;
+        if (span != NULL) {
+          (*nodes)[i]->span_label_start = span->label_start;
+          (*nodes)[i]->span_label_end = span->label_end;
+          (*nodes)[i]->span_value_start = span->value_start;
+          (*nodes)[i]->span_value_end = span->value_end;
+          (*nodes)[i]->span_start = span->span_start;
+          (*nodes)[i]->span_end = span->span_end;
+
+          if (span->filename != NULL || span->filename->str != NULL) {
+              (*nodes)[i]->filename = span->filename->str;
+          }
+          (*nodes)[i]->has_span_info = true;
+        }
+        else {
+          (*nodes)[i]->has_span_info = false;
+        }
+
         i += 1;
     }
     ERR_BAIL(aug);
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -836,6 +836,68 @@ int aug_get(const struct augeas *aug, const char *path, const char **value) {
     return -1;
 }
 
+int aug_get_nodes(const struct augeas *aug, const char *pathin, struct aug_node ***nodes) {
+    struct pathx *p = NULL;
+    struct tree *tree;
+    int cnt = 0;
+
+    api_entry(aug);
+
+    if (nodes != NULL)
+        *nodes = NULL;
+
+    if (STREQ(pathin, "/")) {
+        pathin = "/*";
+    }
+
+    p = pathx_aug_parse(aug, aug->origin, tree_root_ctx(aug), pathin, true);
+    ERR_BAIL(aug);
+
+    for (tree = pathx_first(p); tree != NULL; tree = pathx_next(p)) {
+        if (! TREE_HIDDEN(tree))
+            cnt += 1;
+    }
+    ERR_BAIL(aug);
+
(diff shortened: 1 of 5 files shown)


thedrow:
I have some idea on what this failure is about. `aug_get_nodes()` is not exported to libaugeas.so.0 which is referenced by libaugeas.la.
I have no idea why though

thedrow:
Found the problem. I'm currently debugging the segfault.

thedrow:
@lutter @raphink I must have allocated something wrong because when i reaches 2 the test causes a segfault. Care to take a look?

thedrow:
Sorry about the ping. This works now.

thedrow:
I think we should add span information as well if it is available.

thedrow:
I managed to get all the information in one pass instead of two using a linked list. This will be much more efficient. I also added the span information if it exists.

@raphink @lutter This is ready for review.

lutter:
@thedrow is your comment that using the existing API is inefficient based on performance measurements ? If so, could you share some of those ? I agree that the additional parsing of paths that `aug_get` and `aug_label` do is unnecessary work, but I am not sure how bad that is in practice.

I am a little hesitant about the current patches for a few reasons:
* exposing the struct in the header file will make it very hard to change that struct without breaking API compatibility
* the structs expose data that is allocated by augeas, and could be deallocated (like the label and value); since it's not clear to users when that might happen, they need to process all the `aug_node` data (and free it) before they make any other call to augeas

For these reasons, it would be better to keep this data in opaque structs, but even that won't take away the memory management problems.

One simple way to improve the efficiency of `aug_get`, `aug_label` etc. is to do something like
```
    aug_defvar(aug, "myvar", complex_path_expression);
    char **paths;
    int n = aug_match(aug, "$myvar", &paths);
    for(int i = 1; i <= n; i++) {
       char *px;
       asprintf(&px, "$myvar[%d]", i);
       // Get information about the node
       const char *value, *label;
       aug_get(aug, px, &value);
       aug_label(aug, px, &label);
       printf("path %s = %s (label = %s)\n", paths[i], value, label);
       aug_span(aug, px, ...);
    }
```

I'd be curious to see how performant storing the result of a complex match in a variable is; what happens internally is that all the nodes get collected into a nodeset (whose contents are properly managed when tree nodes get deleted) and accessing the nodeset only involves very simple path expressions which are very fast to evaluate.

If that is too slow, we could add API calls that access nodesets directly, i.e. something like `aug_nodeset_get(aug, "myvar", i, &value)`. This approach would not involve additional allocations, but would be much safer than copying internal data into externally visible data structures.

thedrow:
@lutter Yes. I introduced this API because our existing implementation for OSQuery's augeas table is too slow. Currently `SELECT * FROM augeas` takes about 5 seconds to complete and is peaking at ~95% CPU. This implementation shortens the query runtime to half a second and peaks at 87%-95% CPU.
I'm testing your suggestion right now.

thedrow:
I tested your suggestion. `SELECT * FROM augeas` takes about 202 seconds to complete.
Both master and my branch match exactly the same nodes.

You can check out the code changes I made here: https://github.com/thedrow/osquery/blob/augeas-var-iterator/osquery/tables/system/posix/augeas.cpp

thedrow:
@lutter ping?

lutter:
@thedrow sorry, was on vacation and traveling .. I'll look into this, but it'll take me a few days

lutter:
@thedrow rereading your last comment, I am a bit confused about timing: are you saying your original code (with just straight calls to aug_get etc.) takes 5s, but going through defvar and iterating over that takes 202s ?

thedrow:
@lutter Exactly. I haven't found the root cause for the performance issue yet but using an iterator with defvar is super slow.

lutter:
@thedrow I just posted PR #481. In my testing, this greatly speeds up dumping large numbers of nodes (on my laptop, running `AUGEAS_LENS_LIB=$aug/lenses examples/dump '//*'` takes 40s for the roughly 92k matching nodes, but when I pass `-n` to dump, it only takes about 300ms.

Please have a look and let me know if that addresses your performance concerns. As I said above, I like that better than exposing structs because (1) it's easy to add more `aug_ns_*` functions without breaking backwards compatibility (2) variables are safe even if nodes get deleted.

Also, let me know if you need other `aug_ns_*` functions

thedrow:
I'm closing this PR. We'll continue discussion on #418. 
Thanks.

PR #480, opened 2017-08-05
* src/pathx.c (parse_name): correctly handle trailing whitespace in n…

lutter:
…ames

When a name ended in whitespace, we incorrectly assumed it was always ok to
trim that whitespace. That is not true if that whitespace is escaped,
i.e. if the path expression is something like '/x\ '. In that case, the
name really needs to be literally 'x ', i.e., we can not trim that
whitespace.

The incorrect behavior led to turning '/x\ ' first into 'x\' and then,
because we assume that '\' is always followed by a character inside the
string, when we removed the escaping '\', we would read beyond the end of
the intermediate string result; if we were lucky, that would lead to a
crash, otherwise we'd continue with junk.

We now make sure that escaped whitespace at the end of a string does not
get stripped, avoiding all these headaches.

Fixes RHBZ https://bugzilla.redhat.com/show_bug.cgi?id=1475621

### DIFF
--- a/src/pathx.c
+++ b/src/pathx.c
@@ -1710,6 +1710,16 @@ int pathx_escape_name(const char *in, char **out) {
     return 0;
 }
 
+/* Return true if POS is preceded by an odd number of backslashes, i.e., if
+ * POS is escaped. Stop the search when we get to START */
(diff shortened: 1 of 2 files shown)


carnil:
This is CVE-2017-7555.

PR #481, opened 2017-08-08
WIP: speed up iterating and getting information about lots of nodes

lutter:
This PR introduces new functions `aug_ns_*` which make it possible to iterate quickly over a large number of nodes and retrieve information for them. See `examples/dump.c` for how to use them and a small program that can be used to assess the performance improvement of these functions.

This is still work in progress, as it needs tests and some more thinking about error handling from `aug_ns_*`.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -1915,6 +1915,28 @@ int aug_print(const struct augeas *aug, FILE *out, const char *pathin) {
     return result;
 }
 
+static char *
+tree_source(const augeas *aug, struct tree *tree) {
+    char *result = NULL;
+
+    while (!(ROOT_P(tree) || tree->file))
+        tree = tree->parent;
+
+    if (tree->file) {
+        if (tree->span == NULL) {
+            int r;
+            r = ALLOC(tree->span);
+            ERR_NOMEM(r < 0, aug);
+            tree->span->filename = make_string(path_of_tree(tree));
+            ERR_NOMEM(tree->span->filename == NULL, aug);
+        }
+        result = strdup(tree->span->filename->str);
+        ERR_NOMEM(result == NULL, aug);
+    }
+ error:
+    return result;
+}
+
 int aug_source(const augeas *aug, const char *path, char **file_path) {
     int result = -1, r;
     struct pathx *p = NULL;
@@ -1935,13 +1957,9 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
               r, path);
     ERR_THROW(r == 0, aug, AUG_ENOMATCH, "There is no node matching %s",
               path);
-    while (!(ROOT_P(match) || match->file))
-        match = match->parent;
 
-    if (match->file) {
-        *file_path = path_of_tree(match);
-        ERR_NOMEM(file_path == NULL, aug);
-    }
+    tree_source(aug, match);
+    ERR_BAIL(aug);
 
     result = 0;
  error:
@@ -1990,6 +2008,43 @@ int tree_equal(const struct tree *t1, const struct tree *t2) {
         t2 = t2->next;
     }
     return t1 == t2;
+}
+
+int aug_ns_attr(const augeas* aug, const char *var, int i,
+                const char **value, const char **label, char **file_path) {
+    int result = -1;
+
+    if (value != NULL)
+        *value = NULL;
+
+    if (label != NULL)
+        *label = NULL;
+
+    if (file_path != NULL)
+        *file_path = NULL;
+
+    api_entry(aug);
+
+    struct tree *tree = pathx_symtab_get_tree(aug->symtab, var, i);
+    ERR_THROW(tree == NULL, aug, AUG_ENOMATCH,
+              "Node %s[%d] does not exist", var, i);
+
+    if (file_path != NULL) {
+        *file_path = tree_source(aug, tree);
+        ERR_BAIL(aug);
+    }
+
+    if (value != NULL)
+        *value = tree->value;
+
+    if (label != NULL)
+        *label = tree->label;
+
+    result = 1;
+
+ error:
+    api_exit(aug);
+    return result;
 }
 
 /*
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,11 @@
       back to text fails. They now make it clearer what part of the tree
       was problematic, and what the tree should have looked like.
     * Fixed the pkg-config file, which should now be usable
+  - API changes
+    * add function aug_ns_attr to allow iterating through a nodeset
+      quickly. See examples/dump.c for an example of how to use them
+      instead of aug_get, aug_label etc. and for a way to measure
+      performance gains.
   - Lens changes/additions
     * Sshd: split HostKeyAlgorithms into list of values
 
--- a/configure.ac
+++ b/configure.ac
@@ -65,7 +65,7 @@ if test x"$enable_debug" = x"yes"; then
 fi
 
 dnl Version info in libtool's notation
-AC_SUBST([LIBAUGEAS_VERSION_INFO], [22:0:22])
+AC_SUBST([LIBAUGEAS_VERSION_INFO], [23:0:23])
 AC_SUBST([LIBFA_VERSION_INFO], [5:4:4])
 
 AC_GNU_SOURCE
--- a/examples/Makefile.am
+++ b/examples/Makefile.am
@@ -2,10 +2,15 @@
 GNULIB= ../gnulib/lib/libgnu.la
 GNULIB_CFLAGS= -I $(top_srcdir)/gnulib/lib
 
-AM_CFLAGS = @AUGEAS_CFLAGS@ @WARN_CFLAGS@ $(GNULIB_CFLAGS) \
+AM_CFLAGS = @AUGEAS_CFLAGS@ @WARN_CFLAGS@ @LIBXML_CFLAGS@ $(GNULIB_CFLAGS) \
 			-I $(top_srcdir)/src
 
-bin_PROGRAMS = fadot
+bin_PROGRAMS = fadot dump
 
 fadot_SOURCES = fadot.c
 fadot_LDADD = $(top_builddir)/src/libfa.la $(GNULIB)
+
+dump_sources = dump.c
+dump_LDFLAGS = -static
+dump_LDADD =  $(top_builddir)/src/libaugeas.la $(top_builddir)/src/libfa.la
+	          $(GNULIB)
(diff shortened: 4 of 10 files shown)


src/augeas.c
```diff
@@ -1993,6 +1993,80 @@ int tree_equal(const struct tree *t1, const struct tree *t2) {
 }
 
 /*
+ * aug_ns_* functions
+ */
+int aug_ns_get(augeas* aug, const char *var, int i, const char **value) {
```
thedrow: Shouldn't aug's type be `const augeas* aug`?

src/augeas.c
```diff
@@ -1993,6 +1993,80 @@ int tree_equal(const struct tree *t1, const struct tree *t2) {
...
+        *value = tree->value;
+
+    result = 0;
+
+ error:
+    api_exit(aug);
+    return result;
+}
+
+int aug_ns_label(augeas *aug, const char *var, int i, const char **label) {
```
thedrow: Shouldn't aug's type be `const augeas* aug`?

src/augeas.c
```diff
@@ -1993,6 +1993,80 @@ int tree_equal(const struct tree *t1, const struct tree *t2) {
...
+
+    result = 1;
+
+ error:
+    api_exit(aug);
+    return result;
+}
+
+int aug_ns_source(augeas *aug, const char *var, int i, char **file_path) {
+    int result = -1;
```
thedrow: Shouldn't aug's type be `const augeas* aug`?

src/augeas.c
```diff
@@ -1993,6 +1993,80 @@ int tree_equal(const struct tree *t1, const struct tree *t2) {
...
+    api_entry(aug);
+
+    struct tree *tree = pathx_symtab_get_tree(aug->symtab, var, i);
+    ERR_THROW(tree == NULL, aug, AUG_ENOMATCH,
+              "Node %s[%d] does not exist", var, i);
+
+    if (value != NULL)
+        *value = tree->value;
+
+    result = 0;
```
thedrow: Result doesn't match how aug_get() returns it's result. I expected aug_ns_get to return 1 on success.
It might not make sense but it breaks user expectations.

thedrow:
Using the aug_ns_* methods I'm able to get osquery to select on all nodes in 1.5 seconds. This is still far from the 0.5 seconds it took to do so in #474.
There is an upside though, CPU utilization is averaging at 74.75.
Running the query with these changes keeps utilization under 80%.

I haven't tried the example code as of yet but I will do so shortly.

src/augeas.c
lutter: Yes, totally right. Fixing that.

src/augeas.c
lutter: Yes, totally right. Fixing that.

src/augeas.c
lutter: Yes, totally right. Fixing that.

src/augeas.c
lutter: Yes, totally right. Fixing that to return 1 on success.

lutter:
I've updated the code with your comments, and made one change to improve performance: aug_ns_source (and aug_source) now cache the filename in the span for the toplevel node in a file. On my machine, that brings the time to dump a tree from ~ 280ms down to ~ 80ms.

Give it a spin and let me know how that performs for you.

thedrow:
Ok now we're at 1s which is getting to where we want to be.
As far as I'm concerned this is ready for merge unless you can come up with more optimizations.

thedrow:
Actually I think we can shave a few more cycles by providing an API to get the label, value and file path at once. What do you think?

lutter:
Have you tried that yet ? Would be curious to see how much the speedup is.

Another possible speedup would be to make the `file_path` that `aug_ns_source` returns const, i.e., have the caller not deallocate it as that would save one unnecessary copy operation within Augeas. But that would make `aug_source` (whic hwe can't change) and `aug_ns_source` work differently.

How big is the tree you are dumping ? Can you tell me how long running `dump` on your whole tree takes ? (It's 80ms on my machine; I just want to gauge how much tim your code spends in Augeas internals and how much in your code - if it now only spends a small percentage of its time in Augeas, there's not much point in speeding that part up)

thedrow:
We spend 73% of the time in `aug_match()`. The rest are mainly string operations.
I'd argue that there's still no need to traverse the tree multiple times if you want all the data both for the normal case and for the nodeset case. Regardless of performance, it provides a nice API for the user.

I have 41835 nodes in my tree.
I have disabled the printfs since they were not letting me benchmark correctly.

Time to dump tree with normal match: time: 1251ms -1355ms
Time to dump tree with aug_ns_*: 9ms - 12ms

It's impressive!

I'm able to shave a couple of more milliseconds with an API that gets all the values at once.
Time to dump tree with aug_ns_all: 7ms - 9ms

thedrow:
```patch
diff --git a/examples/dump.c b/examples/dump.c
index 33400aa8..4a5684ef 100644
--- a/examples/dump.c
+++ b/examples/dump.c
@@ -119,6 +119,43 @@ static void dump_var(struct augeas *aug, const char *path) {
     }
 }
 
+/*
+ * Print out information for all nodes matching PATH using aug_ns_all
+ * function
+ */
+static void dump_varall(struct augeas *aug, const char *path) {
+    int nmatches;
+    int i;
+
+    /* Define the variable 'matches' to hold all the nodes we are
+       interested in */
+    aug_defvar(aug, "matches", path);
+
+    /* Count how many nodes we have */
+    nmatches = aug_match(aug, "$matches", NULL);
+    if (nmatches < 0) {
+        fprintf(stderr, "aug_match for '%s' failed\n", path);
+        fprintf(stderr, "error: %s\n", aug_error_message(aug));
+        exit(1);
+    }
+
+    fprintf(stderr, "using var and aug_ns_*\n");
+    fprintf(stderr, "%d matches for %s\n", nmatches, path);
+
+    for (i=0; i < nmatches; i++) {
+        const char *value, *label;
+        char *file = NULL;
+
+        /* Get information about the ith node, equivalent to calling
+         * aug_get etc. for "$matches[i]" but much more efficient internally
+         */
+        aug_ns_all(aug, "matches", i, &value, &label, &file);
+
+        //printf("%d: %s %s %s\n", i, label, value, file);
+        free(file);
+    }
+}
+
 static void print_time_taken(const struct timeval *start,
                              const struct timeval *stop) {
     time_t elapsed = (stop->tv_sec - start->tv_sec)*1000
@@ -129,13 +166,28 @@ static void print_time_taken(const struct timeval *start,
 int main(int argc, char **argv) {
     int opt;
     int use_var = 0;
+    int use_varall = 0;
     const char *pattern = "/files//*";
     struct timeval stop, start;
-
-    while ((opt = getopt(argc, argv, "n")) != -1) {
+    //
+    // while ((opt = getopt(argc, argv, "n")) != -1) {
+    //     switch (opt) {
+    //     case 'n':
+    //         use_var = 1;
+    //         break;
+    //     default:
+    //         fprintf(stderr, "Usage: %s [-n] [pattern]\n", argv[0]);
+    //         fprintf(stderr, "      without '-n', iterate matches\n");
+    //         fprintf(stderr, "      with '-n', use a variable and aug_ns_*\n");
+    //         exit(EXIT_FAILURE);
+    //         break;
+    //     }
+    // }
+
+    while ((opt = getopt(argc, argv, "a")) != -1) {
         switch (opt) {
-        case 'n':
-            use_var = 1;
+        case 'a':
+            use_varall = 1;
             break;
         default:
             fprintf(stderr, "Usage: %s [-n] [pattern]\n", argv[0]);
@@ -154,6 +206,9 @@ int main(int argc, char **argv) {
     gettimeofday(&start, NULL);
     if (use_var) {
         dump_var(aug, pattern);
+    }
+    else if(use_varall) {
+        dump_varall(aug, pattern);
     } else {
         dump_match(aug, pattern);
     }
diff --git a/src/augeas.c b/src/augeas.c
index 0fec6f3f..7412fa80 100644
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -2079,6 +2079,42 @@ int aug_ns_source(const augeas *aug, const char *var, int i, char **file_path) {
     return result;
 }
 
+int aug_ns_all(const augeas *aug, const char *var, int i,
+               const char **value, const char **label, char **file_path) {
+  int result = -1;
+
+  api_entry(aug);
+
+  ARG_CHECK(file_path == NULL, aug,
+            "aug_ns_source: FILE_PATH must not be NULL");
+
+  if (value != NULL)
+      *value = NULL;
+
+  if (label != NULL)
+      *label = NULL;
+
+  *file_path = NULL;
+
+  struct tree *tree = pathx_symtab_get_tree(aug->symtab, var, i);
+  ERR_THROW(tree == NULL, aug, AUG_ENOMATCH,
+            "Node %s[%d] does not exist", var, i);
+
+  if (label != NULL)
+      *label = tree->label;
+
+  if (value != NULL)
+      *value = tree->value;
+
+  *file_path = tree_source(aug, tree);
+  ERR_BAIL(aug);
+
+  result = 0;
+error:
+  api_exit(aug);
+  return result;
+}
+
 /*
  * Error reporting API
  */
diff --git a/src/augeas.h b/src/augeas.h
index f9e6b227..8a1832aa 100644
--- a/src/augeas.h
+++ b/src/augeas.h
@@ -529,6 +529,9 @@ int aug_ns_label(const augeas *aug, const char *var, int i, const char **label);
  */
 int aug_ns_source(const augeas *aug, const char *var, int i, char **file_path);
 
+int aug_ns_all(const augeas *aug, const char *var, int i,
+               const char **value, const char **label, char **file_path);
+
 /*
  * Error reporting
  */
diff --git a/src/augeas_sym.version b/src/augeas_sym.version
index d9b42ab4..d67c7a43 100644
--- a/src/augeas_sym.version
+++ b/src/augeas_sym.version
@@ -86,4 +86,5 @@ AUGEAS_0.23.0 {
       aug_ns_get;
       aug_ns_label;
       aug_ns_source;
+      aug_ns_all;
 } AUGEAS_0.22.0;
```

This is the patch for the code that adds `aug_ns_all()` to the API. I can add it after we finish this PR or you can modify it to your liking if you want to.

lutter:
I've folded those three functions into one `aug_ns_attr` and added some tests. If this looks good to you and works for you, I'd be happy to merge it.

thedrow:
LGTM. Can we get an official release with this API?

lutter:
@thedrow I'll try and get a release out by the end of the week, but am not sure if that's actually going to happen

thedrow:
Awesome!!!

PR #483, opened 2017-08-18
Add FreeBSD paths

bitnexus:
Adds FreeBSD support for the Postfix lenses.

I noticed that other lenses don't change the comments (eg. `About: Configuration files`) to reflect the additional paths, so I didn't either. Glad to modify if needed.


bitnexus:
Added a few more commits for different paths missing for FreeBSD. Let me know if I should make separate PRs for this kind of thing.

lutter:
Looks good to me ! Thanks for doing this.

PR #488, opened 2017-08-24
Fix sudoers lens: recognize "match_group_by_gid"

### LINKED ISSUE #482: Missing new options for Sudoers.aug
In RHEL7.4 the latest version of sudo (1.8.1p2-10) contains new option(s) which are not captured in the constructed options list in Sudoers.aug. I'm not sure if there are others but 'match_group_by_gid' is one that causes the default /etc/sudoers file to fail parsing.

ltoscano-rh:
The option is now enabled by default in the default sudoers of
RHEL 7.4 (and probably soon CentOS 7).

Closes #482


lutter:
Thanks so much for your contribution !

PR #489, opened 2017-08-27
Added branch prediction macros and marked allocation errors as unlikely

thedrow:
There are probably a lot more places where these macros should be used but this is a good start.

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -138,6 +138,26 @@
 #endif
 
 /**
+* AUGEAS_LIKELY:
+*
+* Macro to flag a code branch as a likely branch
+*
+* AUGEAS_UNLIKELY:
+*
+* Macro to flag a code branch as an unlikely branch
+*/
+#ifndef __has_builtin
+#   define __has_builtin(x) (0)
+#endif
+#if defined(__builtin_expect) || __has_builtin(__builtin_expect)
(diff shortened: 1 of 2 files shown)


thedrow:
@lutter This is ready for review :)

src/internal.h
```diff
@@ -138,6 +138,26 @@
...
+*
+* AUGEAS_UNLIKELY:
+*
+* Macro to flag a code branch as an unlikely branch
+*/
+#ifndef __has_builtin
+#   define __has_builtin(x) (0)
+#endif
+#if defined(__builtin_expect) || __has_builtin(__builtin_expect)
+#   define AUGEAS_LIKELY(x)        (__builtin_expect(!!(x), 1))
```
lutter: Could we just call them `LIKELY` and `UNLIKELY` or is that going to clash with some other include ?

lutter:
I like the idea .. out of curiosity, has that made a perf difference for you ?

src/internal.h
thedrow: These macros are always prefixed by the project name since other projects may also define their own set of builtin_expect macros.

thedrow:
I don't see any measurable performance improvement but than again I haven't done anything other than running the dump example.
In any case, this is a good idea to tell the CPU that error branches are rarely executed.

lutter:
Merged as f98b04cc

PR #491, opened 2017-09-04
Cgconfig: allow fperm & dperm in admin & task

ptoscano:
These keys are used to control the permissions for files and directories.


lutter:
Looks good. Thanks for the PR !

PR #494, opened 2017-09-11
Grub: handle top-level "boot" directive

ptoscano:
Grub 1 effectively ignores commands in the configuration which work only in the command line.  The generated configuration by anaconda included also a commented `boot=device` entry at the beginning: uncommenting that does not make the configuration invalid, but makes the Grub lens not able to parse it.

Since there is no harm in representing a configuration key that will be effectively ignored, accept top-level `boot` entries as well.


raphink:
LGTM

PR #498, opened 2017-09-19
Fix memory leak when checking version of augparse

qiankehan:
Add aug_close() before exit to fix memory leak when printing
version information of augparse.


lutter:
Nice ! Thanks for the patch !

PR #499, opened 2017-09-19
A few portability fixes

smortex:
This PR fixes a few portability issues found when trying to hack some code from a FreeBSD host.


lutter:
Yes, these are great ! Thanks so much for the PR

PR #500, opened 2017-09-20
* .gitignore: ignore examples/dump

ptoscano:


lutter:
Thanks for the patch !

PR #501, opened 2017-09-20
Add seccomp_sandbox to vsftpd lens

stroebs:
Required for Docker containers. Enabled by default on amd64
https://wiki.gentoo.org/wiki/Vsftpd
seccomp filter sanboxing with vsftpd 3.0.x


lutter:
Thanks for the patch !

PR #502, opened 2017-09-25
More branch prediction

thedrow:
Same as #489 just more of it :)


lutter:
To repeat what I said in IRC: when building with `CFLAGS='-O3 -g'`, at least `examples/dump -n` gets a little slower on my machine with this patch (~35ms to dump 85k nodes vs ~ 30ms without this patch)

So, unless there's some other performance tests that shows that this improves things, I'd not merge this.

lutter:
Closing this because of inactivity. If there's anything new regarding this, feel free to reopen.

PR #503, opened 2017-09-27
Fix several httpd issues

lutter:


lutter:
@SwartzCr @joohoi I would love to get your feedback on this - could you just run your tests,and replace your `httpd.aug` with the one from this PR ? I mostly want to make sure I ma not introducing any indavertent regressions

joohoi:
I just tested this as a drop-in replacement without any other changes. The amount of errors and test failures I'm seeing suggests that there must be more going on than just simple lens errors (42 errors and 13 failures). Is there a minimum augeas version requirement for the updated lens? I'm running libaugeas version 1.8.0 (from Debian Stretch; libaugeas0_1.8.0-1+deb9u1_amd64)

lutter:
@joohoi yes, you are right - I inadvertently had a dependency on commit 244c0edd, which made the lens not work with augeas 1.8.0 or earlier. I've now changed things so that it will work with 1.8.0. I'd appreciate if you could give this another spin and let me know how that goes.

joohoi:
All tests seem to pass with at least somewhat recent version of libaugeas: https://travis-ci.org/certbot/certbot/builds/280566778

I'll have to test it further with older distributions (and hence older libaugeas versions) we're supporting.

lutter:
That's quite encouraging. I just dropped the new httpd lens and tests into a build of augeas-1.0.0 and those tests passed, too. But please let me know if your testing turns up any issues.

lutter:
@joohoi I'll merge this - since you vendor your own version of Httpd, even if there should be a problem with these changes, they won't immediately affect you, and we can always fix them as they come up.

PR #513, opened 2017-10-09
Implemented fa_export function

pevalme:
I have implemented a function called "fa_export" which basically outputs the FA as an array such that each position corresponds to a possible transitions. Whenever there is a 1, the transitions exists in the FA.

```
v[i * num_states * ALPHABET_SIZE + j *ALPHABET_SIZE + k] = 1 iff transition i→j is labelled with k
```

The goal of this function is to allow a translation from the current FA structure to the one used by any application that uses libfa. Before this change, the only way to export the FA goes through outputting it as a .dot file.

### DIFF
--- a/src/fa.c
+++ b/src/fa.c
@@ -4513,6 +4513,93 @@ void fa_dot(FILE *out, struct fa *fa) {
     fprintf(out, "}\n");
 }
 
+void fa_json(FILE *out, struct fa *fa) {
+    int *list_hashes = NULL;
+    int list_size = 100;
+    int num_states = 0;
+    int it;
+    char first = true;
+
+    fprintf(out,"{\n\t\"final\": [");
+
+    F(ALLOC_N(list_hashes, list_size));
+    // list_hashes = (int *)malloc(sizeof(int) * list_size);
+
+    list_for_each(s, fa->initial) {
+        if (num_states == list_size - 1){
+            list_size += list_size;
+            F(REALLOC_N(list_hashes, list_size));
+            // list_hashes = realloc(list_hashes, list_size);
+        }
+        // Store hash value
+        list_hashes[num_states] = s->hash;
+        // We use the hashes to map states to Z_{num_states}
+        s->hash = num_states++;
+        if (s->accept) {
+            if (!first) fprintf(out, ", %ld", s->hash);
+            else fprintf(out,"%ld", s->hash);
+            first = false;
+        }
+    }
+
+    fprintf(out, "],\n\t\"deterministic\": %d,\n\t\"transitions\": [\n", fa->deterministic ? 1 : 0);
+
+    first = true;
+    list_for_each(s, fa->initial) {
+        for_each_trans(t, s) {
+            if (!first) fprintf(out, ",\n");
+            first = false; 
+            fprintf(out, "\t\t{ \"from\": %ld, \"to\": %ld, \"on\": \"",s->hash, t->to->hash);
+            print_char(out, t->min);
+            if (t->min != t->max) {
+                fputc('-', out);
+                print_char(out, t->max);
+            }
+            fprintf(out, "\" }");
+        }
+    }
+
+    fprintf(out,"\n\t]\n}");
+
+error:
+    // Restoring hash values to leave the FA structure untouched
+    it = 0;
+    list_for_each(s, fa->initial) {
+        if (it == num_states) break;
+        s->hash = list_hashes[it++];
+    }
+    free(list_hashes);
+}
+
+int fa_is_deterministic(struct fa *fa) {
+    return fa->deterministic;
+}
+
+struct state *fa_state_initial(struct fa *fa) {
+    return fa->initial;
+}
+
+bool fa_state_is_accepting(struct state *st) {
+    return st->accept;
+}
+
+struct state* fa_state_next(struct state *st) {
+    return st->next;
+}
+
+size_t fa_state_num_trans(struct state *st) {
+    return st->tused;
+}
+
+bool fa_state_trans(struct state *st, size_t i, struct state **to, unsigned char *min, unsigned char *max) {
+    if (st->tused <= i) return -1;
+
+    (*to) = st->trans[i].to;
+    (*min) = st->trans[i].min;
+    (*max) = st->trans[i].max;
+    return 0;
+}
+
 /*
  * Local variables:
  *  indent-tabs-mode: nil
--- a/src/fa.h
+++ b/src/fa.h
@@ -25,9 +25,14 @@
 
 #include <stdio.h>
 #include <regex.h>
+#include <stdbool.h>
+#include <stdint.h>
 
 /* The type for a finite automaton. */
 struct fa;
+
+/* The type of a state of a finite automaton */
+struct state;
 
 /* Denote some basic automata, used by fa_is_basic and fa_make_basic */
 enum fa_basic {
@@ -271,6 +276,34 @@ int fa_expand_nocase(const char *regexp, size_t regexp_len,
  * memory, and -2 if FA has more than LIMIT words.
  */
 int fa_enumerate(struct fa *fa, int limit, char ***words);
+
+/* Print FA to OUT as a JSON file. State 0 is always the initial one */
+void fa_json(FILE *out, struct fa *fa);
+
+/* Returns 1 if the FA is deterministic and 0 otherwise */
+int fa_is_deterministic(struct fa *fa);
+
+/* Return the initial state */
+struct state *fa_state_initial(struct fa *fa);
+
+/* Return true if this is an accepting state */
+bool fa_state_is_accepting(struct state *st);
+
+/* Return the next state; return NULL if there are no more states */
+struct state* fa_state_next(struct state *st);
+
+/* Return the number of transitions for a state */
+size_t fa_state_num_trans(struct state *st);
+
+/* Produce details about the i-th transition. 
+ *
+ * On success, *to points to the destination state of the transition and
+ * the interval [min-max] is the label of the transition.
+ * 
+ * On failure, *to, min and max are not modified.
+ *
+ * Return 0 on success and -1 otherwise */
+bool fa_state_trans(struct state *st, size_t i, struct state **to, unsigned char *min, unsigned char *max);
 
 #endif
 
--- a/examples/fadot.c
+++ b/examples/fadot.c
@@ -31,8 +31,11 @@
 #include <string.h>
 #include <getopt.h>
 #include <stdlib.h>
+#include <limits.h>
 
 #include "fa.h"
+
+#define UCHAR_NUM (UCHAR_MAX+1)
 
 const char *progname;
 
@@ -45,8 +48,8 @@ static void usage(void) {
   fprintf(stderr, "\nCompile REGEXP and apply operation to them. By default, just print\n");
   fprintf(stderr, "the minimized regexp.\n");
   fprintf(stderr, "\nOptions:\n\n");
-  fprintf(stderr, "  -o OPERATION       one of : show concat union intersect\n");
-  fprintf(stderr, "                              complement minus example\n");
+  fprintf(stderr, "  -o OPERATION       one of : show concat union intersect json\n");
+  fprintf(stderr, "                              complement minus example print\n");
   fprintf(stderr, "  -f DOT_FILE        Path of output .dot file\n");
   fprintf(stderr, "  -n                 do not minimize resulting finite automaton\n");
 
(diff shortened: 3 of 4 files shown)


lutter:
Hi @pevalme ! Can you explain a little bit more about how `fa_export` would be used and what it is good for ? I fear that turning an FA into a `num_states * num_states * ALPHABET_SIZE` will be very inefficient, but that depends on how this information will be used. Can you also explain more about what you mean by `the format that any application that uses libfa` means ? The function `fa_as_regexp` makes it possible to turn an FA into a regexp so that any tool that understands POSIX ERE can use them.

One alternative to exporting this as a (possibly very large) array of char would be to add functions that allow interrogating the structure of the FA. Or to more directly add the operations you need to perform on them to the libfa API.

pevalme:
Right now I'm working on a project in which I need to build an automaton from a regular expression for what I want to use your library. However, once the automaton is built I need to store it in a particular format so that my program works efficiently. (For instance I need to store some extra information for each symbol that labels an edge and I want instant access to it. Basically I'll have an array of symbols and each symbol contains a list of transitions)
I could have built a function that transforms your format into mine but that would result in a very specific transformation and it won't make any sense to add it to the library. This is why I decided to implement function `fa_export` that uses this naive format to output the automaton so my program can now process it.

Adding functions to query the FA is going to be very complex since we have to allow many different operations. However, it might be worth to modify the format of the output produced by `fa_export` to produce:
* A list of states, each with a list of outgoing transitions, or
* A list of symbols, each with a list of edges labeled with it.

If you find these transformations useful I can implement both of them.

lutter:
I really like the idea of making it possible to annotate FA's with additional information; I am wondering if it wouldn't be best to have `fa_export` produce output that is easy to use with a wide variety of tools, for example, to export to JSON and define a format for that. The advantage of that would be that you can easily read that from lots of languages, and we would be able to add more information if that is ever needed without breaking backwards compatibility. What I am thinking of is that `fa_export` would either produce a string, or print to a `FILE*` in the following format:

```json
{
  "final": [ 7, 12, 132],
  "transitions": [
    { "from": 1, "to": 12, "on": "abcv" },
    { "from": 1, "to": 7, "on": "u" },
    { "from": 7, "to": 12, "on": "\n" }
  ]
}
```

The format would work as follows:
* the initial state is always state 1 (or should this start at state 0 ?)
* states are just integers, and are defined just by mentioning them
* the `final` array lists which states are final
* the `transitions` are an array, where each entry specifies `from` which state `to` which state the transition goes, and `on` what characters (a string that just acts as a byte array, though for JSON, byte values above `0x7f` would need to be UTF-8 encoded

There are variations on this that might work, too, for example something that follows the internal representation more closely, e.g., list states, and for each state list transitions by giving the `to` and the characters `on` which the transition occurs.

What do you think about going that way ?

BTW, the `re` field in transitions is only used when `fa_as_regexp` is called; that works by eliminating intermediate states one-by-one, replacing their transitions with regular expressions. IOW, not something to worry about when dealing with FA's outside of that function.

pevalme:
After the changes I've just done in this pull request, the library features the following new functions:
* `fa_json`: This function writes down in a given file a JSON description of the FA. It follows the format described in the previous comment.
* `fa_export_bytearray`: This function encodes the FA in a `char` pointer received as argument (which is allocated within the function, thus the argument is a pointer to pointer) according to the description provided in my previous comment.
* `fa_export_list`: This function encodes the FA according to the following structure, which is a simplification of the one used internally by the library.
```C
struct FA_EXPORT {
    int state; // Current state
    char final; // Is final?
    struct FA_EXPORT *next; // Next state in the list
    int num_trans; // Number of outgoing transitions of the current state
    struct EDGE *trans; // [destination state, min label, max label]
};

struct EDGE {
    int end, min, max;
};
In this case the FA is not manipulated so it may have several accepting states. This data structure is public and can be handled by the program using the library.
```
* `fa_export_list_free`: Since the user of the library does not know how the memory is handled, I added this function to free the memory used by a `struct FA_EXPORT`.
-----
The goal of functions `fa_export_list` and `fa_export_bytearray` is to allow the user of the library to manipulate a copy of the FA without the need for writing it into a file and then reading it.

lutter:
Hi @pevalme,

wow .. thanks for doing all this ! If we have an `fa_export` function, do we need all the other functions in the library though ? With the structure in hand, users can easily write their own JSON exporter (or bytearray encoder)

I'll leave some more detailed comments on the files.

src/fa.h
```diff
@@ -53,6 +53,19 @@ enum fa_minimization_algorithms {
  */
 extern int fa_minimization_algorithm;
 
+/* Data structure defining the exported automata as a list*/
+struct FA_EXPORT {
+    int state; // Current state
```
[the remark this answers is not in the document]
pevalme: Removing state number implies modifying the edges so for each transition we store a pointer to the destiny rather than an identifier.
Using these numbers reduces the memory usage and, in my opinion, makes the structure easier to manipulate.

Anyways, I will change the variable type.

lutter:
@pevalme I am sorry if I was not clear in my last comment - my question was if adding the interfaces I outlined in that comment instead of the ones you suggested in your patches would serve your needs. It seems that with the fa_state_* functions, you could implement fa_export etc. in your own code. The reason why I'd prefer that is that these will be easier to maintain in the long term, and pose less of a risk of introducing problems.

I'd be totally fine with adding functionality to print the fa with the 'export' and 'json' operations to fadot, just that they'd work in terms of the fa_state_* functions. It would also be better to move the code in fadot that does the printing into separate functions to keep the code a bit more readable.

pevalme:
Alright, I totally misunderstood your previous comment. 
Anyways, @lutter, I think it would be easier to use my structure
```
struct fa_export {
    bool final; // Is final?
    struct fa_export *next; // Next state in the list
    uint32_t num_trans; // Number of outgoing transitions of the current state
    struct fa_edge *trans; // [destination state, min label, max label]
};

struct fa_edge {
    struct fa_export *end;
    uint32_t min, max;
};
```
and function `int fa_export_list` to allow the user go through the automaton, rather than using all your functions.
At the end of the day, using the `fa_state_*` functions requires the `fa` to follow the `fa_export` structure (list of states where each member has a list of transitions) and the user needs to know it. At the end of the day I don't see why providing these functions results in a more general/easier to maintain approach.

Sorry if I'm missing something obvious.

lutter:
The difference with doing it with the functions vs doing it via a struct is ABI stability: if the layout of the structs ever needs to be changed, that breaks the ABI and requires that applications using them are recompiled. Using the functions doesn't have the same drawback: if there is additional data that needs to be exposed, we can just add a new `fa_state_*` function to retrieve that and old code just keeps working without a rebuild. That's why, from a maintenance perspective, I'd really want to expose this functionality through the `fa_state_*` functions (and you can easily reuse the code you already have for your own purposes with a fairly small edit to change it from direct access to `fa` structs to using the `fa_state_*` functions)

src/fa.c
```diff
@@ -4513,6 +4513,88 @@ void fa_dot(FILE *out, struct fa *fa) {
...
+
+    list_for_each(s, fa->initial) {
+        if (num_states == list_size - 1){
+            list_size += list_size;
+            list_hashes = realloc(list_hashes, list_size);
+        }
+        // Store hash value
+        list_hashes[num_states] = s->hash;
+        // We use the hashes to map states to Z_{num_states}
+        s->hash = num_states++;
```
lutter: Nice ! I like how this just reuses `s->hash` to number the states.

src/fa.c
```diff
@@ -4513,6 +4513,88 @@ void fa_dot(FILE *out, struct fa *fa) {
...
+            if (t->min != t->max) {
+                fputc('-', out);
+                print_char(out, t->max);
+            }
+            fprintf(out, "\" }");
+        }
+    }
+
+    fprintf(out,"\n\t]\n}");
+
```
lutter: This is probably where the `error` label should go (so that we restore hashes on error)

src/fa.c
```diff
@@ -4513,6 +4513,88 @@ void fa_dot(FILE *out, struct fa *fa) {
...
+    // Restoring hash values to leave the FA structure untouched
+    it = 0;
+    list_for_each(s, fa->initial) {
+         s->hash = list_hashes[it++];
+    }
```
lutter: You also need to `free(list_hashes)` so we don't leak that.

src/fa.c
```diff
@@ -4513,6 +4513,88 @@ void fa_dot(FILE *out, struct fa *fa) {
     fprintf(out, "}\n");
 }
 
+void fa_json(FILE *out, struct fa *fa) {
+    int *list_hashes;
```
lutter: Initialize this to `NULL`. We won't hit the error label right now before we try to allocate, but it makes it safer in case we ever change the code so that it could encounter an error before allocation.

src/fa.h
```diff
@@ -25,10 +25,15 @@
...
 #include <stdio.h>
 #include <regex.h>
+#include <stdbool.h>
+#include <stdint.h>
 
 /* The type for a finite automaton. */
 struct fa;
 
+/* The type of a state of a finite automaton */
+struct state;
```
lutter: This should carry a dire warning that pointers to struct state are only valid across calls to `fa_state_*`. If you get a `struct state *` and call any other `fa_*` function, you must consider the pointer invalid (since we could have garbage collected the state as a result of that call)

src/fa.h
```diff
@@ -272,6 +277,27 @@ int fa_expand_nocase(const char *regexp, size_t regexp_len,
  */
 int fa_enumerate(struct fa *fa, int limit, char ***words);
 
+/* Print FA to OUT as a JSON file. State 0 is always the initial one */
+void fa_json(FILE *out, struct fa *fa);
+
+/* Returns 1 if the FA is deterministic and 0 otherwise */
+int fa_is_deterministic(struct fa *fa);
```
lutter: Should probably return a `bool`

src/fa.h
```diff
@@ -272,6 +277,27 @@ int fa_expand_nocase(const char *regexp, size_t regexp_len,
...
+
+/* Return the initial state */
+struct state *fa_state_initial(struct fa *fa);
+
+/* Return true if this is an accepting state */
+int fa_state_is_accepting(struct state *st);
```
lutter: Should be `bool`

src/fa.h
```diff
@@ -272,6 +277,27 @@ int fa_expand_nocase(const char *regexp, size_t regexp_len,
...
+
+/* Return the next state; return NULL if there are no more states */
+struct state* fa_state_next(struct state *st);
+
+/* Return the number of transitions for a state */
+uint32_t fa_state_num_trans(struct state *st);
```
lutter: The `uint32_t` should be `size_t`

src/fa.h
```diff
@@ -272,6 +277,27 @@ int fa_expand_nocase(const char *regexp, size_t regexp_len,
...
+
+/* Produce details about the i-th transition; return 0 on success, or -1 of there is no such transition */
+int fa_state_trans(struct state *st, uint32_t i, struct state **to, unsigned char *min, unsigned char *max);
```
lutter: The `uint32_t` should be `size_t`. Mention in the comment that `to`, `min`, and `max` must not be `NULL`, and none of them will have been modified if `fa_state_trans` returns -1 (or maybe change the behavior to set `*to = NULL` in `fa_state_trans`).

src/fa.c
```diff
@@ -4513,6 +4513,88 @@ void fa_dot(FILE *out, struct fa *fa) {
...
+    int list_size = 100;
+    int num_states = 0;
+    int it;
+    char first = true;
+
+    fprintf(out,"{\n\t\"final\": [");
+
+    list_hashes = (int *)malloc(sizeof(int) * list_size);
```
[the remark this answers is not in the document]
pevalme: If I understand correctly, `F(ALLOC_N(list_hashes, list_size));` will automatically jump to the `error:` label in case of failure. Therefore, when restoring the hashes we need to check that we are not restoring any "false hash" (non-initialized position in the array). I've added the check `if (it == num_states) break;` in the loop to solve this issue.

src/fa.h
pevalme: I don't really know what you mean. Calling any **not** `fa_state*` function with a `struct state *` will cause an error at compilation time.

lutter:
Thanks so much for doing this work and for your patience ! I have now merged this as 40210763.

I've trtied to explain what I meant by 'do not use a struct state* across calls to other fa_*' functions in this comment in fa.h:
```c
/* The type of a state of a finite automaton. The fa_state functions return
 * pointers to this struct. Those pointers are only valid as long as the
 * only fa_* functions that are called are fa_state_* functions. For
 * example, the following code will almost certainly result in a crash (or
 * worse):
 *
 *     struct state *s = fa_state_initial(fa);
 *     fa_minimize(fa);
 *     // Crashes as S will likely have been freed
 *     s = fa_state_next(s)
 */
struct state;
```

Does that make it clearer ?

In any event, thanks again for doing this !

src/fa.c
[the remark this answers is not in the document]
lutter: Yes, good catch - I've turned that into an if statement around the loop, but you are absolutely right that we need to make sure we do not try to switch hashes back in if we never changed them.

Issue #486, opened 2017-08-22
Syslog lens : include support


rmdir:
Some syslog servers have an include directive. It makes augeas parser failed

Eg.
```
/augeas/files/etc/syslog.conf/error = "parse_failed"
/augeas/files/etc/syslog.conf/error/pos = "1488"
/augeas/files/etc/syslog.conf/error/line = "37"
/augeas/files/etc/syslog.conf/error/char = "0"
/augeas/files/etc/syslog.conf/error/lens = "/usr/local/share/augeas/lenses/dist/syslog.aug:258.18-.51:"
/augeas/files/etc/syslog.conf/error/message = "Get did not match entire input"
```
```
nl /etc/syslog.conf |grep 37
    37  include						/etc/syslog.d
```

domcleal:
Which server supports this? We tend to use per-application lenses if possible.

rmdir:
@domcleal It's include in freebsd's syslogd since r308160 
https://svnweb.freebsd.org/base?view=revision&revision=308160

PR #521, opened 2017-10-28
Syslog: accept include directive

lutter:
Add syslog.conf from
https://svnweb.freebsd.org/base?view=revision&revision=308160 to test root,
too

Fixes https://github.com/hercules-team/augeas/issues/486


lutter:
@rmdir could you have a look and verify that that's what you had in mind in #486 ?

smortex:
Hello!

There is [interest in merging this into the FreeBSD port of augeas](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=232200) so that we ship this feature on FreeBSD.  However, we currently cannot use this PR patch because of conflicts with the `NEWS` file.

@lutter, would you mind refactoring a bit this pull request by splitting it in two commits:

* One including the actual fix (changes to `lenses/syslog.aug`, `lenses/tests/test_syslog.aug` and `tests/root/etc/syslog.conf`);
* One that updates the `NEWS` file.

We could then directly use the first commit patch in the port, ad you would be able to sync the `NEWS` file if desired without impact on the port we produce.

Thanks!

raphink:
Merged as c608a4b6cd8b655325492ebc7dc2f3cc06de3252

lutter:
@smortex is having this merged into master enough for you ? It'll show up in the next Augeas release anyway; if you want it earlier, you can just grab the updated lens and install that in the default lens path after Augeas' `make install`

smortex:
@lutter sure, it's good for me! 

Thanks!

PR #524, opened 2017-11-04
Support all YAML multiline modes

jayvdb:
Only `|-` was supported.
Added all variations of `[|>][-+]?`

### DIFF
--- a/lenses/yaml.aug
+++ b/lenses/yaml.aug
@@ -28,8 +28,18 @@ View: indent
 *)
 let indent = del /[ \t]+/ "  "
 
-let mval = [ label "@mval" . Util.del_str "|-" . eol
-           . [ label "@line" . indent . store Rx.space_in . eol ]+ ]
+let mline = [ label "@line" . indent . store Rx.space_in . eol ]+
+
+let m_literal_clip = [ label "@mlitclip" . Util.del_str "|" . eol . mline ]
+let m_literal_strip = [ label "@mval" . Util.del_str "|-" . eol . mline ]
+let m_literal_keep = [ label "@mlitkeep" . Util.del_str "|+" . eol . mline ]
+
+let m_fold_clip = [ label "@mfoldclip" . Util.del_str ">" . eol . mline ]
+let m_fold_strip = [ label "@mfoldstrip" . Util.del_str ">-" . eol . mline ]
+let m_fold_keep = [ label "@mfoldkeep" . Util.del_str ">+" . eol . mline ]
+
+let mval = (m_literal_clip | m_literal_strip | m_literal_keep
+           | m_fold_clip | m_fold_strip | m_fold_keep)
 
 (*
 View: inherit
--- a/lenses/tests/test_yaml.aug
+++ b/lenses/tests/test_yaml.aug
@@ -73,3 +73,11 @@ test YAML.lns get "abc:
       { "@mval"
         { "@line" = "ghi" } } } }
 
+test YAML.lns get "abc:
+  def: |
+  ghi
+\n" =
+  { "abc"
+    { "def"
+      { "@mlitclip"
+        { "@line" = "ghi" } } } }


jayvdb:
No, this is bad.  See https://github.com/hercules-team/augeas/issues/525

jayvdb:
Well, it doesnt make the situation worse.

lenses/yaml.aug
```diff
@@ -28,8 +28,18 @@ View: indent
...
-           . [ label "@line" . indent . store Rx.space_in . eol ]+ ]
+let mline = [ label "@line" . indent . store Rx.space_in . eol ]+
+
+let m_literal_clip = [ label "@mlitclip" . Util.del_str "|" . eol . mline ]
+let m_literal_strip = [ label "@mval" . Util.del_str "|-" . eol . mline ]
+let m_literal_keep = [ label "@mlitkeep" . Util.del_str "|+" . eol . mline ]
+
+let m_fold_clip = [ label "@mfoldclip" . Util.del_str ">" . eol . mline ]
+let m_fold_strip = [ label "@mfoldstrip" . Util.del_str ">-" . eol . mline ]
+let m_fold_keep = [ label "@mfoldkeep" . Util.del_str ">+" . eol . mline ]
```
lutter: Instead og having all these different labels, how about doing something like
```
let mval = [ label "@mval" . store /[|>][+-]?/ . eol . mline ]
```

lutter:
Thanks for doing this; I am more than happy to merge a variation of this, though have to warn you that full YAML support will probably not be possible with Augeas in any event. At least, I don't know of a way how YAML's indentation could be turned into the proper nesting/recursion in the Augeas tree.

jayvdb:
Yea, I wasted a weekend trying to get better support ... and got a headache for my effort, and started implementing an indent counter into the language but ran out of time .  Is there a bug about this deficiency in the language?  Ongoing design discussion?  Does it affect any other important formats for augeas?

I'm happy to tidy up this patch incorporating review comments.

lenses/yaml.aug
jayvdb: I'm very green at augeas, so I'll be happily led by others.

I was trying to avoid storing syntactical symbols in the augeas result.  labels looked like they could be a nicer way of showing the meaning for end-users.
Is there a way to have two labels?  `[|>]` and `[+-]` say two different things.

lutter:
The problem is unfortunately somewhat of a math problem: 'normal' Augeas lenses can only deal with file formats that are regular languages; using recursive (`let rec`) lenses also lets you describe context-free languages. YAML, because of the significance of the indentation, is neither, and instead is a context-sensitive language, and therefore out of the realm of what can be described with Augeas.

The best way to fix it would be to make it possible to preprocess a file: you'd turn indentation into markers that say 'indentation level increased' resp. 'indentation level decreased' - it would basically turn something like
```
a:
  x: v1
  y: v2
b: v3
```

into 

```
{a: { x: v1 } { y: v2 } } {b: v3 }
```
where the `{` and `}` are the special indendation marker. That transform would turn it into a context-free language, which could be described with a lens similar to `Json.lns`. The transform has to be, of course, bidirectional; you'll also have to go from the representation on the bottom to the one on the top.

All that would be a significant undertaking, and would require surgery in the Augeas implementation itself - basically, right now, Augeas knows that it has to read the contents of a file, hand that to a lens and get a tree back (and vice versa). To support YAML, you'd have to add a transformation step between reading the file and handing it to the lens, and the lens language would need a way to say that that should happen - I think there's no point in trying to describe the actual transformation in the lens language, and you'd rather just implement them in C.

jayvdb:
I guess your reply means there are no good answers to my offtopic questions above, and the next step is a strategy/design discussion on the mailing list about formats not yet possible to support with augeas.  It is beyond the scope of this PR.

On topic is https://github.com/hercules-team/augeas/pull/524#discussion_r150372885 , which hasnt been answered yet.

lutter:
Sorry for dropping the ball on this .. there's no way to have multiple labels for a node (the tree is incredibly simple: each node has one label and one value)

But it would be possible to store whether multiline strings get clipped/stripped/kept in the value of the node. Have a look at [this branch](https://github.com/lutter/augeas/tree/pr/524) where I implemented that and let me know how useful you think that is.

lutter:
Closing this since there hasn't been any activity for quite some time. Feel free to reopen if there is anything more that can be done.

PR #526, opened 2017-11-05
Support YAML sequence documents

jayvdb:
This only adds support for a document which
consists of a sequence.


lutter:
Looks good. Thanks for the patch !

PR #527, opened 2017-11-05
spelling fixes

ka7:
patch contains some spelling fixes ( just in comments ) as found by a bot ( http://www.misfix.org, https://github.com/ka7/misspell_fixer ). 

( I fixed also the `man/augtool.1` which might be in dispute, as it's supposed to be re-generated from the .pod. But as it's already hanging around here... )


lutter:
Awesome ! Thanks !

PR #533, opened 2017-12-11
Update Makefile.am

joerg-krause:
Fix missing trailing backslash for `dump_LDADD`.


lutter:
Thanks for catching this ! Can't believe this slipped through ...

PR #537, opened 2018-01-05
httpd: include  /etc/httpd/conf.modules.d

examon:
Fedora/RHEL uses this directory for modules config.


lutter:
Makes perfect sense. Thanks for the patch !

PR #539, opened 2018-01-11
Add logind.conf to systemd lense

jcpunk:
logind.conf is also a ini file style config provided by systemd


lutter:
Thanks for the patch !

PR #541, opened 2018-02-01
Cleaner error message on regex /

jcpunk:
Since a key can't have an '/' in it due to use as a value seperator for tree elements, I've added that to the error message so folks are fully aware.


lutter:
Thanks for the patch !

PR #546, opened 2018-02-18
Fix parsing of append / prepend / default / supersede statements

jwm:
#96 added support for append/prepend/default/supersede statements, but reused `stmt_hash`, which includes `stmt_hash_re` in its matching.

This means the lens requires one of the keywords `send` or `option` as the first argument to these statements:

```prepend option domain-name-servers 127.0.0.1;```

instead of the [correct syntax](https://www.isc.org/wp-content/uploads/2017/08/dhcp41clientconf.html), which does not allow them:

```prepend domain-name-servers 127.0.0.1;```

### DIFF
--- a/lenses/dhclient.aug
+++ b/lenses/dhclient.aug
@@ -93,19 +93,23 @@ let stmt_array        = [ key stmt_array_re
 let stmt_hash_re      = "send"
                       | "option"
 
-let stmt_hash         = [ key stmt_hash_re
-                        . sep_spc
-                        . ( [ key word . sep_spc . sto_to_spc_noeval ]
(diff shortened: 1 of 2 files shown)


lutter:
Thanks for opening this PR. It's currently failing in Travis. It looks like the problem is that `augparse lenses/tests/test_dhclient.aug` is failing. Could you look into addressing that and updating the PR ?

jwm:
@lutter `make check` is successful now - sorry for the long delay. 😄

lutter:
I did some minor editing of the commit messages and NEWS and merged this as ba04c9a0. Thanks for the patch !

PR #549, opened 2018-03-09
Chrony: add new options

mlichvar:
Add new options supported in chrony 3.2 and 3.3.


lutter:
Thanks for the patch. I just saw that it also updates `.gnulib` - is that needed or could that be dropped from this PR ?

mlichvar:
That wasn't supposed to be there. It should be fixed now. Thanks.

lutter:
Awesome ! Thanks again for the contribution.

PR #560, opened 2018-05-29
Parse nginx sites-enabled

plumbeo:
On Debian/Ubuntu the active sites are configured in /etc/nginx/sites-enabled, either as files or as symlinks to files in /etc/nginx/sites-available/, so parse that path too.

### DIFF
--- a/lenses/nginx.aug
+++ b/lenses/nginx.aug
@@ -120,9 +120,11 @@ let lns = ( Util.comment | Util.empty | directive )*
 let filter = incl "/etc/nginx/nginx.conf"
            . incl "/etc/nginx/conf.d/*.conf"
            . incl "/etc/nginx/sites-available/*"
+           . incl "/etc/nginx/sites-enabled/*"
            . incl "/usr/portage/www-servers/nginx/files/nginx.conf"
            . incl "/usr/local/etc/nginx/nginx.conf"
            . incl "/usr/local/etc/nginx/conf.d/*.conf"
            . incl "/usr/local/etc/nginx/sites-available/*"
+           . incl "/usr/local/etc/nginx/sites-enabled/*"
 
 let xfm = transform lns filter


lutter:
Thanks for the patch !

kingdonb:
Can anyone verify this works? I am new to augeas but have just compiled from master branch, and I can't get any signs of life out of this use case. (It's exactly what I needed, if it does what it says on the tin!)

```
augtool> ls /augeas/load/Nginx
lens = @Nginx
incl[1] = /usr/local/etc/nginx/sites-enabled/*
incl[2] = /usr/local/etc/nginx/sites-available/*
incl[3] = /usr/local/etc/nginx/conf.d/*.conf
incl[4] = /usr/local/etc/nginx/nginx.conf
incl[5] = /usr/portage/www-servers/nginx/files/nginx.conf
incl[6] = /etc/nginx/sites-enabled/*
incl[7] = /etc/nginx/sites-available/*
incl[8] = /etc/nginx/conf.d/*.conf
incl[9] = /etc/nginx/nginx.conf
```

great! But

```
augtool> get /files/etc/nginx/sites-available
/files/etc/nginx/sites-available (o)
```

I wasn't able to find any docs that explain the meaning of `(o)` but I guess it's this guy, putting his hands up in frustration: 🤷‍♂️

domcleal:
> I wasn't able to find any docs that explain the meaning of `(o)`

It means the node doesn't exist. I'd expect to see `(none)` if something was beneath that node, which would indicate the node exists and has no value.

`errors` may give you information about any parsing issues, while `print /files/etc/nginx` may give a clearer, recursive view of what's there, while `get` will only return the value of a single node. Obviously double check you have files in that path, it's an unusual (older?) one.

If you continue to have problems, it'd be best to create a new GitHub issue, ask in #augeas (FreeNode) or on augeas-devel for more help rather than this old PR.

kingdonb:
Thanks, I will head over to there! I didn't see any tests on this PR and it looked like it was merged without comment, so I wasn't sure if it was really a complete work or not. I saw it mentioned in the 1.11.0 release notes which is how I wound up here.

raphink:
I'll add to @domcleal that there's an `#augeas` channel on the [Puppet community slack](http://puppetcommunity.slack.com) (as I confess I'm hardly ever on IRC these days).

PR #562, opened 2018-06-04
Grub: tolerate some invalid entries

lutter:
Refusing to parse an entire file because of invalid entries is too
harsh. Try to make the behavior a little friendlier by just mapping invalid
entries to '#error' nodes but still parsing the rest of the file.


rwmjones:
Yes this looks good in my testing.

Issue #566, opened 2018-06-12
nginx: Support semicolons in values


rgm3:
I would like to use augtool to parse an nginx.conf file that contains:

```
add_header X-XSS-Protection "1; mode=block" always;
```

Unfortunately, I have not as yet been able to appropriately modify the nginx.aug lens to redefine the statement terminator as "semicolons at the end of the line that are not within double quotes."

lutter:
Have a look at PR #567 - that should do what you want. Please let me know if that works for you. 

I am not familiar with Nginx syntax - if there are other kinds of directives that need this treatment, please let me know, too.

lutter:
I've changed the PR so that the lens now accepts simple directives like 'internal'; there's no whitelist for such entries though, it'll accept 'joey;' just the same as a directive.

PR #567, opened 2018-06-13
Nginx: allow semicolons inside double quoted strings in simple direct…

lutter:
…ives

The 'argument' to a simple directive can now be made up of any number of
bare words (not containing quotes) and quoted strings. Semicolons are
allowed within quoted strings.

Fixes https://github.com/hercules-team/augeas/issues/566


rgm3:
This fixes my use case perfectly, although it won't parse 100% of all nginx configs (since nginx also supports backslash escaped chars and other things and doesn't actually require quotes to use semicolons), but that's okay IMO.  This is a much appreciated incremental improvement.

PR #574, opened 2018-07-19
augtool: fix access to invalid memory

ptoscano:
When stripping the context from the result, `readline_path_generator` used to realloc the string to a shorter size, copying only the content after the prefix.  This resulted in reading with `strcpy` from the previous memory, which is freed already.  Avoid the issue, and simplify the code by using `strdup`, freeing the old string.

This issue could be reproduced in `augtool`, trying to autocomplete files without the `/files` prefix, e.g.:
``
augtool> ls <TAB><TAB>
``


lutter:
Thanks for the patch ! Yes, this is much cleaner and safer.

PR #575, opened 2018-07-19
fa: check result of fopen

ptoscano:
Return earlier if `fopen` fails, otherwise `fa_dot` will dereference a null pointer.


lutter:
Nice catch ! Thanks for the patch.

PR #576, opened 2018-07-20
Fix few minor memory leaks

ptoscano:
Fix a few memory leaks, mostly in error situations, or debugging.

Also fix the run of `make check` in the `valgrind` target.


lutter:
Thanks so much for these fixes ! I really appreciate the work and diligence that goes into tracking this stuff down.

Issue #569, opened 2018-06-27
Memory usage issues with large Apache httpd configuration files


joohoi:
An user reported an memory usage issue in Certbot, and when investigating the cause I found out that our (very suboptimal) regexes caused the Augeas `match` memory usage to balloon out of proportions.

This only happens with really large configurations, and complex regexes however. The reporting users configuration had roughly 300 configuration files with 1500 lines each. I was able to create a much simplified file with similar effect. The effects start to be really visible after 40k lines, and at 80k lines `augtool` with stock httpd lens segfaults in startup on Debian stretch.

Please note that the regex is really suboptimal, and is used for backwards compatibility for environments with too old version of Augeas to support case insensitive regex flag `'i'`. Using recent version and the case insensitive flag reduces the memory footprint roughly by two thirds, but it is still roughly 400 fold compared to actual configuration size.

## How to reproduce

1) Modify the following template, repeating the line:  `AddType huge/memory       usage	` 50k times. 
```
<VirtualHost *:80>
	ServerName memory.issue
	ServerAdmin webmaster@localhost
	DocumentRoot /var/www/html
	<Directory /var/www/html>
		<IfModule mod_mime.c>
			# Repeat the following line 50k times
			AddType huge/memory       usage	
		</IfModule>
	</Directory>
</VirtualHost>
```

2) Copy the modified template to `/etc/apache2/sites-available/memorybomb.conf`

3) Start `augtool -I httpd.aug`

4) Use suboptimal regex: 
```
match /files/etc/apache2/sites-available/*[label()=~regexp('.*\\.conf')]//*[self::directive=~regexp('([Ii][Nn][Cc][Ll][Uu][Dd][Ee])|([Ii][Nn][Cc][Ll][Uu][Dd][Ee])|([Ii][Nn][Cc][Ll][Uu][Dd][Ee][Oo][Pp][Tt][Ii][Oo][Nn][Aa][Ll])')]
```

5) Observe the memory usage, it should use roughly 2Gb of memory during the regex match, roughly 1000 times the on-disk size of the configuration.

lutter:
Thanks for the detailed description. I am having trouble reproducing this behavior. I followed your instructions, and with a file where the line is repeated 50k times, memory usage only goes up to ~ 120MB.

It also doesn't seem to make a difference whether I use that match or not in terms of memory usage.

[...]

PR #578, opened 2018-08-09
Replace pure function invocations in path expressions with their result

lutter:
In path expressions, we generally need to evaluate functions against every
node that we consider for the result set. For example, in the path
expression /files/etc/hosts/*[ipaddr =~ regexp('127\\.')], the regexp
function was evaluated against every entry in /etc/hosts. Evaluating that
function requires the construction and compilation of a new regexp. Because
of how memory is managed during evaluation of path expressions, the memory
used by all these copies of the same regexp is only freed after we are done
evaluating the path expression. This causes unacceptable memory usage in
large files (see https://github.com/hercules-team/augeas/issues/569)

To avoid these issues, we now distinguish between pure and impure functions
in the path expression interpreter. When we encounter a pure function, we
change the AST for the path expression so that the function invocation is
replaced with the result of invoking the function. With the example above,
that means we only construct and compile the regexp '127\\.' once,
regardless of how many nodes it gets checked against. That leads to a
dramatic reduction in the memory required to evaluate path expressions with
such constructs against large files.

Fixes https://github.com/hercules-team/augeas/issues/569

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,8 @@
   - General changes/additions
     * augmatch: add a --quiet option; make the exit status useful to tell
                 whether there was a match or not
+    * Drastically reduce the amount of memory needed to evaluate complex
+      path expressions against large files (Issue #569)
   - API changes
     * aug_source did not in fact return the source; and always returned
       NULL for that. That has been fixed.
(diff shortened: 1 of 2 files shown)


lutter:
Note for posterity: I also tried replacing the pooled memory management scheme for `struct value` used in the path expression interpreter with straightup reference counting for values. That didn't make much of a difference in terms of memory usage, even for very large files.

PR #580, opened 2018-08-10
Redis: accept bind command

lutter:
Fixes https://github.com/hercules-team/augeas/issues/194

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -20,6 +20,8 @@
              allow semicolons inside double quoted strings in
              simple directives, and allow simple directives without
              an argument (Issue #566)
+    * Redis: accept the 'bind' statement with multiple IP addresses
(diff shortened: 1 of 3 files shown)


lutter:
Slight rework of PR #197

lutter:
The difference between this and PR #197 is that one bind statement `bind ip1 ip2` now creates a tree
```
  { "bind" { "ip" = ip1 } { "ip" = ip2 } }
```

The extra `ip` in there is necessary so that we can distinguish between one bind statement with two IP addresses and two bind statements with one IP address each. They need to generate different trees so that we know what to write out when saving the tree.

Using `match /files/etc/redis/redis.conf/bind/ip` will list all the IP's from all bind statements.

PR #585, opened 2018-10-08
lexer: disable 'input' and 'yyunput' functions

ptoscano:
flex by default generates the `input`, and `yyunput` functions; OTOH, augeas does not use them, so add the right options to suppress their generation.


lutter:
Nice catch! Thanks for the patch.

PR #586, opened 2018-10-10
DevfsRules: Add lens for FreeBSD devfs.rules files

raphink:
This is for @igalic to review

### DIFF
--- a/lenses/devfsrules.aug
+++ b/lenses/devfsrules.aug
@@ -14,7 +14,7 @@ module DevfsRules =
             . Util.del_str "]" . eol
             . counter "entry"
 
-  let record = IniFile.record title entry
+  let record = IniFile.record title (entry | comment)
 
   let lns = IniFile.lns record comment
 
--- a/lenses/devfsrules.aug
+++ b/lenses/devfsrules.aug
@@ -1,0 +1,24 @@
+module DevfsRules =
+
+  autoload xfm
+
+  let comment  = IniFile.comment IniFile.comment_re "#"
+
+  let eol = Util.eol
+
+  let line_re = /[^][#; \t\n][^#;\n]*[^#; \t\n]/
+  let entry = [ seq "entry" . store line_re . (eol | comment) ]
+
+  let title = Util.del_str "["
+            . key Rx.word . [ label "id" . Sep.equal . store Rx.integer ]
+            . Util.del_str "]" . eol
+            . counter "entry"
+
+  let record = IniFile.record title entry
+
+  let lns = IniFile.lns record comment
+
+  let filter = incl "/etc/defaults/devfs.rules"
+            .  incl "/etc/devfs.rules"
+
+  let xfm = transform lns filter
--- a/lenses/tests/test_devfsrules.aug
+++ b/lenses/tests/test_devfsrules.aug
@@ -1,0 +1,59 @@
+module Test_DevfsRules =
+
+  let manpage_example = "[localrules=10]
+add path 'da*s*' mode 0660 group usb
+"
+
+  test DevfsRules.lns get manpage_example =
+  { "localrules" { "id" = "10" }
+      { "1" = "add path 'da*s*' mode 0660 group usb" } }
+
+
+  let example = "[devfsrules_jail_unhide_usb_printer_and_scanner=30]
+add include $devfsrules_hide_all
+add include $devfsrules_unhide_basic
+add include $devfsrules_unhide_login
+add path 'ulpt*' mode 0660 group printscan unhide 
+add path 'unlpt*' mode 0660 group printscan unhide
+add path 'ugen2.8' mode 0660 group printscan unhide  # Scanner (ugen2.8 is a symlink to usb/2.8.0)
+add path usb unhide
+add path usbctl unhide
+add path 'usb/2.8.0' mode 0660 group printscan unhide
+
+[devfsrules_jail_unhide_usb_scanner_only=30]
+add include $devfsrules_hide_all
+add include $devfsrules_unhide_basic
+add include $devfsrules_unhide_login
+add path 'ugen2.8' mode 0660 group scan unhide  # Scanner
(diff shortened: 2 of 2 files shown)


igalic:
EXCITING

lenses/devfsrules.aug
```diff
@@ -0,0 +1,24 @@
+module DevfsRules =
+
+  autoload xfm
+
+  let comment  = IniFile.comment IniFile.comment_re "#"
+
+  let eol = Util.eol
+
+  let line_re = /[^][#; \t\n][^#;\n]*[^#; \t\n]/
+  let entry = [ seq "entry" . store line_re . (eol | comment) ]
```
igalic: can you explain this one?

can we add a negative test?

raphink: Lines don't really have a key/value pattern, it's more entire lines to store, so I'm using a `seq` approach for that, with a counter called "entry", which is reset in the `title` definition so that entries are numbered inside each section. We then store `line_re`, which is anything but characters that will clash with `eol` or `comment`.

raphink:
Does that work for you @igalic?

igalic:
it does indeed! — from `augparse`, at least — it parses all my rules so far

i just seem to have forgotten how to sensibly use augeas from puppet
especially: how to add numbered items?! will the `id` be incremented automatically on set?

igalic:
> it does indeed! — from `augparse`, at least — it parses all my rules so far

***my*** rules… funnily enough, [`/etc/defaults/devfs.rules`](https://gist.github.com/b8a157c12f42fd4bd04810ccccebf189) is not parsed:

```sh-session
root@container-host1 ~# augtool -l /etc/defaults/devfs.rules 
augtool> ls /
augeas/ = (none)
files = (none)
augtool> ls files
augtool> 
```

raphink:
@igalic 	if it doesn't parse, type `errors` in the `augtool` session to check for parsing errors.

igalic:
why didn't i think of that? — perhaps because it doesn't tell me there's an error >_>

```sh-session
root@container-host1 ~# augtool -l /etc/defaults/devfs.rules 
augtool> errors
Error in /etc/defaults/devfs.rules:25.0 (parse_failed)
  Get did not match entire input
  Lens: /usr/local/share/augeas/lenses/dist/inifile.aug:497.25-.43:
augtool> 
```

raphink:
Can you post your `devfs.rules` file so I can see what's wrong?

igalic:
https://gist.github.com/b8a157c12f42fd4bd04810ccccebf189 i did

raphink:
OK, I'll check. For the record, you can use https://github.com/raphink/augeas-sandbox/blob/master/augcheck to debug that:

```
$ augcheck /tmp/devfs.rules DevfsRules -I lenses/
/tmp/augcheck.YiS4qP/parse_file.aug:3.0-.62:exception thrown in test
/tmp/augcheck.YiS4qP/parse_file.aug:3.5-.58:exception: Get did not match entire input
    Lens: lenses/inifile.aug:497.25-.43:
    Error encountered at 25:0 (969 characters into string)
    <rules_hide_all=1]\nadd hide\n\n|=|# Basic devices typically ne>

    Tree generated so far:
    /#comment[1] = "The following are some default rules for devfs(5) mounts."
/#comment[2] = "The format is very simple. Empty lines and lines beginning"
/#comment[3] = "with a hash '#' are ignored. If the hash mark occurs anywhere"
/#comment[4] = "other than the beginning of a line, it and any subsequent"
/#comment[5] = "characters will be ignored.  A line in between brackets '[]'"
/#comment[6] = "denotes the beginning of a ruleset. In the brackets should"
/#comment[7] = "be a name for the rule and its ruleset number. Any other lines"
/#comment[8] = "will be considered to be the 'action' part of a rule"
/#comment[9] = "passed to the devfs(8) command. These will be passed"
/#comment[10] = "\"as-is\" to the devfs(8) command with the exception that"
/#comment[11] = "any references to other rulesets will be expanded first. These"
/#comment[12] = "references must include a dollar sign '$' in front of the"
/#comment[13] = "name to be expanded properly."
/#comment[14] = "$FreeBSD: releng/11.2/etc/defaults/devfs.rules 233048 2012-03-16 21:30:26Z mm $"
/#comment[15] = "Very basic and secure ruleset: Hide everything."
/#comment[16] = "Used as a basis for other rules."
/devfsrules_hide_all
/devfsrules_hide_all/id = "1"
/devfsrules_hide_all/1 = "add hide"
/devfsrules_hide_all/\(none\)


Failed to load /tmp/augcheck.YiS4qP/parse_file.aug
```

which shows where the lens fails to parse.

raphink:
@igalic 	this should do

igalic:
i think there's some deep misunderstanding going on here:

```sh-session
root@container-host1 ~# augtool -l /etc/devfs.rules 
augtool> ls /
augeas/ = (none)
files/ = (none)
augtool> ls .
etc/ = (none)
augtool> ls etc/
devfs.rules/ = (none)
augtool> ls etc/devfs.rules/
#comment = for DHCP server
bpf_jail/ = (none)
iocage_auto_6/ = (none)
augtool> ls etc/devfs.rules/iocage_auto_6/
id = 6
1 = add include $devfsrules_hide_all
2 = add include $devfsrules_unhide_basic
3 = add include $devfsrules_unhide_login
4 = add path zfs unhide
5 = add path 'bpf*' unhide
augtool> 
```

raphink:
Ah yes, I should've seen that. `/augeas/context` defaults to `/files`, not `/`, so when you do `ls files`, it does `ls /files/files` :wink:

igalic:
🤦‍♀️ forgive me… i'm all rusty… (but not yet fluent in rust!)
so this works fine now: 👍

lutter:
I touched this up a tad bit (adding the requisite entry to `tests/Makefile.am`) and merged it as 42a40453

PR #593, opened 2018-11-28
Fixes for rsyslog and xorg.conf

arnolda:
These are two small fixes for the rsyslog and xorg.conf lenses.

In rsyslog.conf, #!, #+ and #- don't have a special meaning as they have in syslog, and therefore the usual Util.comment can and should be used.

 In xorg.conf, values for options can be empty strings. This is seldom used, but there are use cases. In our case, we use an empty fbdev to keep autoprobing drivers, but disabling the fbdev driver. Another use case are Xkb layout options.


raphink:
Thanks for the PR. For the sake of tracking features, could you separate this into 2 PRs, one for each lense, please?

arnolda:
Sure, just opened two new pull requests and closing this one.

PR #595, opened 2018-11-30
Rsyslog: bsd-like #!/+/- are just comments without special meaning

arnolda:
In rsyslog.conf, #!, #+ and #- don't have a special meaning as they have in syslog, and therefore the usual Util.comment can and should be used.


lutter:
Thanks for the patch!

PR #596, opened 2018-11-30
Xorg: Accept empty values for options

arnolda:
In xorg.conf, values for options can be empty strings. This is seldom used, but there are use cases. In our case, we use an empty fbdev to keep autoprobing drivers, but disabling the fbdev driver. Another use case are Xkb layout options.


lutter:
Thanks for the patch !

PR #599, opened 2018-12-21
Permit ! in dovecot block titles

### LINKED ISSUE #598: Dovecot lens doesn't allow ! in block titles
Hi,

In fairly recent Dovecot versions, ! is allowed at the start of block names, and in fact is used in the default config:

```
protocol !indexer-worker {
  # If folder vsize calculation requires opening more than this many mails from
  # disk (i.e. mail sizes aren't in cache already), return failure and finish
  # the calculation via indexer process. Disabled by default. This setting must
  # be 0 for indexer-worker processes.
  #mail_vsize_bg_after_count = 0
}
```

The current lens fails on this. I've got a fix, and will open a pull request with it in a moment.

nward:
Fixes #598

### DIFF
--- a/lenses/dovecot.aug
+++ b/lenses/dovecot.aug
@@ -114,7 +114,7 @@ let block_newlines (entry:lens) (comment:lens) =
 Map block enclosed in brackets recursively.
 Block may be indented and have optional argument.
 Block body may have entries, comments, empty lines, and nested blocks recursively. *)
-let rec block = [ indent . key block_names . (Sep.space . Quote.do_dquote_opt (store /[\/A-Za-z0-9_-]+/))? . block_newlines (entry|block|mailbox) comment . eol ]
+let rec block = [ indent . key block_names . (Sep.space . Quote.do_dquote_opt (store /!?[\/A-Za-z0-9_-]+/))? . block_newlines (entry|block|mailbox) comment . eol ]
 
 
 (******************************************************************
--- a/lenses/tests/test_dovecot.aug
+++ b/lenses/tests/test_dovecot.aug
@@ -523,6 +523,10 @@ mail_attachment_dir =
 mail_attachment_min_size = 128k
 mail_attachment_fs = sis posix
 mail_attachment_hash = %{sha1}
+
+protocol !indexer-worker {
+  mail_vsize_bg_after_count = 0
+}
 "
 test Dovecot.lns get mail_conf =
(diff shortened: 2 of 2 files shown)


lenses/dovecot.aug
```diff
@@ -114,7 +114,7 @@ let block_newlines (entry:lens) (comment:lens) =
 Map block enclosed in brackets recursively.
 Block may be indented and have optional argument.
 Block body may have entries, comments, empty lines, and nested blocks recursively. *)
-let rec block = [ indent . key block_names . (Sep.space . Quote.do_dquote_opt (store /[\/A-Za-z0-9_-]+/))? . block_newlines (entry|block|mailbox) comment . eol ]
+let rec block = [ indent . key block_names . (Sep.space . Quote.do_dquote_opt (store /!?[\/A-Za-z0-9_-]+/))? . block_newlines (entry|block|mailbox) comment . eol ]
```
raphink: Is the `!` really allowed inside the quotes?

nward: Sorry - I'd missed this comment.

Yes it is - either inside or outside has the same behaviour. I'm sorry I'm not familiar with the dovecot config parser code enough to examine the code - but testing seems to indicate that it's OK.

I don't know how we could permit "inside or outside" here.

raphink: OK. Could you provide some examples of valid syntax in the test file?

lutter:
@nward are you still interested in this? If so, please add some (simple) examples of the new behavior to `lenses/tests/test_dovecot.aug` Having those tests in there will make sure we don't inadvertently break the behavior you just added.

nward:
I am, yes. Sorry, have been bogged down with a deployment (which, incidentally, uses this patch).
Let me have a look and see if I can figure that out..

lutter:
Cool. Let me know if you run into issues with the tests. Even if the tests are just `test Dovecot.lns get "CRAZY WILD STRING\n" = ?` would be good. I can figure out the rest, I am just not a Dovecot expert and would need the examples to better understand what needs to be checked (and the examples should be valid config snippets).

BTW, you can run individual tests with `augparse lenses/tests/test_dovecot.aug`

nward:
I've pushed a commit which adds a test - and it passes with `make check` (and if I revert my original commit my test fails, so things look right) but it does not pass with `augparse lenses/tests/test_dovecot.aug` and specifically references the test changes I made.. I'm not too sure what the story is..

Issue #444, opened 2017-02-16
shellvars: Syntax error


josephcsible:
```
/augeas/files/etc/profile.d/vte.sh/error = "parse_failed"
/augeas/files/etc/profile.d/vte.sh/error/pos = "970"
/augeas/files/etc/profile.d/vte.sh/error/line = "26"
/augeas/files/etc/profile.d/vte.sh/error/char = "18"
/augeas/files/etc/profile.d/vte.sh/error/lens = "lenses/shellvars.aug:248.12-.60:"
/augeas/files/etc/profile.d/vte.sh/error/message = "Syntax error"
```
Here's the contents of that file:
```
# Copyright © 2006 Shaun McCance <shaunm@gnome.org>
# Copyright © 2013 Peter De Wachter <pdewacht@gmail.com>
#
# This program is free software: you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation, either version 3 of the License, or
# (at your option) any later version.
#
# This program is distributed in the hope that it will be useful,
# but WITHOUT ANY WARRANTY; without even the implied warranty of
# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
# GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License
# along with this program.  If not, see <http://www.gnu.org/licenses/>.

# Not bash or zsh?
[ -n "$BASH_VERSION" -o -n "$ZSH_VERSION" ] || return 0

# Not an interactive shell?
[[ $- == *i* ]] || return 0

# Not running under vte?
[ "${VTE_VERSION:-0}" -ge 3405 ] || return 0

__vte_urlencode() (
  # This is important to make sure string manipulation is handled
  # byte-by-byte.
  LC_ALL=C
  str="$1"
  while [ -n "$str" ]; do
    safe="${str%%[!a-zA-Z0-9/:_\.\-\!\'\(\)~]*}"
    printf "%s" "$safe"
    str="${str#"$safe"}"
    if [ -n "$str" ]; then
      printf "%%%02X" "'$str"
      str="${str#?}"
    fi
  done
)

# Print a warning so that anyone who's added this manually to his PS1 can adapt.
# The function will be removed in a later version.
__vte_ps1() {
  echo -n "(__vte_ps1 is obsolete)"
}

__vte_osc7 () {
[...]

PR #600, opened 2019-01-07
Few simple Shellvars improvements

ptoscano:
Few small improvements for the `Shellvars` lens:
- accept functions wrapped in round brackets, instead of curly brackets (part of #444)
- accept variables with a dash in their name
- exclude csh/tcsh profile scripts


raphink:
Great work!

PR #607, opened 2019-02-22
Ssh: accept RekeyLimit

lutter:
@mboldi-uptycs this should address your issue. I fixed it slightly differently from how you had it in your bug report. Let me know if that works for you.


mboldi-uptycs:
OK  TY

PR #610, opened 2019-02-26
Make the examples standalone

ptoscano:
Make sure the examples can be built standalone, even outside the augeas builds.

See the commit message of the first patch to get more details about this.


lutter:
Merged as commit 8602598e

PR #611, opened 2019-03-07
Fixes missing recognition of double quoted filenames.

wornet-mwo:
Logrotate is able use quote filenames. This by itself is fixed there. Some
distributions files are using that. Logrotate would also be able to use files
with spaces (when quoted), fixing is too complicated for a quick fix and this
should be also uncommon in real life.

### DIFF
--- a/lenses/logrotate.aug
+++ b/lenses/logrotate.aug
@@ -19,7 +19,7 @@ module Logrotate =
    let eol = Util.eol
    let num = Rx.relinteger
    let word = /[^,#= \n\t{}]+/
-   let filename = /\/[^,#= \n\t{}]+/
+   let filename = /(\/[^,#= \n\t{}]+|"\/[^,#= \n\t{}]+")/
    let size = num . /[kMG]?/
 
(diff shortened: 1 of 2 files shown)


lenses/logrotate.aug
```diff
@@ -19,7 +19,7 @@ module Logrotate =
    let eol = Util.eol
    let num = Rx.relinteger
    let word = /[^,#= \n\t{}]+/
-   let filename = /\/[^,#= \n\t{}]+/
+   let filename = /(\/[^,#= \n\t{}]+|"\/[^,#= \n\t{}]+")/
```
raphink: You could also use `Quote.do_dquote_opt`:

```suggestion
let filename = Quote.do_quote_opt (store /\/[^"',#= \n\t{}]+/)
```

and change `store filename` in `rule` to `filename`

wornet-mwo: Thank you for this hint - this looks better. Added the proposal and fixed the test (the proposed variant does not show quotes as value).

PR #614, opened 2019-03-24
Strongswan: parse lists

kunkku:


lutter:
Merged as commit f342b445

PR #618, opened 2019-04-04
Support unix sockets as server address (fix #617)

### LINKED ISSUE #617: Nginx : upstream block not supported
When I have a [upstream](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#upstream) block in a nginx conf file, it is not supported by Augeas.

The block is:
```
upstream php-handler {
    server unix:/var/run/php/php7.3-fpm.sock;
}
```

Tested on v1.8.0 (Debian) and v1.11.0 (alpine)

raphink:
@Salokyn does that solve it for you?

### DIFF
--- a/lenses/nginx.aug
+++ b/lenses/nginx.aug
@@ -56,8 +56,9 @@ let simple =
 (* View: server
      A simple server entry *)
 let server =
-  [ Util.indent . label "@server" . Util.del_str "server"
-  . [ Sep.space . label "@address" . store word ]
+  let address = /[A-Za-z0-9_.:\/-]+/
+  in [ Util.indent . label "@server" . Util.del_str "server"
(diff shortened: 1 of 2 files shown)


Salokyn:
It seems ok :

```shell
augtool print /files/etc/nginx/conf.d
/files/etc/nginx/conf.d
/files/etc/nginx/conf.d/php-handler.conf
/files/etc/nginx/conf.d/php-handler.conf/upstream
/files/etc/nginx/conf.d/php-handler.conf/upstream/#name = "php-handler"
/files/etc/nginx/conf.d/php-handler.conf/upstream/@server
/files/etc/nginx/conf.d/php-handler.conf/upstream/@server/@address = "unix:/var/run/php/php7.3-fpm.sock"
```

PR #619, opened 2019-04-04
Puppetfile: allow symbols as (optional) values (fix #427)

### LINKED ISSUE #427: Puppetfile lens can't parse "mod 'puppetlabs/apache', :latest"
Hello,

A module can be described as (cf https://docs.puppet.com/pe/latest/cmgmt_puppetfile.html) :
```
mod 'puppetlabs/apache', :latest
```

The problem I have is the portion of code in the lens which matches the version doesn't allow anything like that (colons are excluded from the match): 
```
let version = [ label "@version" . Quote.do_quote (store /[^:\n]+/) ]
```

It doesn't match either the opts because they are of the form "key => value":
```
let opt = [ Util.del_str ":" . key Rx.word . del /[ \t]*=>[ \t]*/ " => "
               . Quote.do_quote (store /[^,\n]*/) ]
```

I tried the following and it works but given it's my first time dealing with augeas, I don't find that really clean nor do I know if it's the best way (not sure how to best "match" latest):
```
let version = [ label "@version" . (Quote.do_quote (store /[^:\n]+/) | 
               (Util.del_str ":" . store /latest/)) ]
```

Any help would be appreciated, I'd have no problem creating a merge request to fix that.

Cheers,
Mathieu

raphink:


raphink:
@mchouque @vchepkov this should fix the issues described in #427

vchepkov:
it does for me, thank you

PR #625, opened 2019-04-23
Dockerfile to build a Augeas image

### LINKED ISSUE #623: What about making an official augeas docker image
Here is a Dockerfile I wrote:

```Dockerfile
FROM alpine

RUN apk add --no-cache libxml2 readline

RUN set -xe \
  ; apk add --no-cache --virtual=.build g++ make bison flex readline-dev libxml2-dev git automake autoconf libtool pkgconf coreutils \
 && git clone git://github.com/hercules-team/augeas \
 && cd augeas \
 && ./autogen.sh && make && make install \
 && cd - \
  ; rm -fr /augeas \
  ; apk del .build

CMD augtool
```

The image weights 20 MB

Salokyn:
Closes #623

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,6 +1,8 @@
 1.13.0 - XXXX-XX-XX
 
   - Lens changes/additions
+    * Grunb: support '+' in kernel command line option names
+             (Pino Toscano) (Issue #647)
     * Mke2fs: parse more common entries between [defaults] and the tags
               in [fs_types], fix the type of few entries, handle the [options]
               stanza (Pino Toscano) (Issue #642)


docker/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM alpine
+
+RUN apk add --no-cache libgcc libxml2 readline bash
+
+RUN set -xe ; \
+    apk add --no-cache --virtual=.build gcc libc-dev make bison flex readline-dev libxml2-dev git automake autoconf libtool pkgconf coreutils \
+ && git clone git://github.com/hercules-team/augeas \
```
raphink: Why clone it instead of copying from the local dir?

Salokyn: I was asking myself the same question. Is it better to build the image with the dockerfile alone or with the full project?

I can change it to copy from the local dir.

PR #629, opened 2019-05-15
* src/put.c (split_concat): initialize regs earlier

ptoscano:
Recent changes (commit 4702d306fe01391af82dc1c7e2e1d50f571fda86) in this helper function cause the code after the `error` label to be used earlier than when `regs` is initialized; since the code in `error` frees a couple of the fields of the `regs` variable, then initialize `regs` earlier to avoid potential freeing uninitialized memory.

Found by Coverity.


ptoscano:
ping

ptoscano:
ping -- @lutter ?

PR #630, opened 2019-05-16
Krb5: improve [dbmodules] and includes

ptoscano:
Improve the way `[dbmodules]` is parsed, and allow `include`/`includedir` everywhere.

### DIFF
--- a/lenses/krb5.aug
+++ b/lenses/krb5.aug
@@ -130,10 +130,19 @@ let dbdefaults =
     simple_section "dbdefaults" keys
 
 let dbmodules =
-  let keys = /db_library|ldap_kerberos_container_dn|ldap_kdc_dn/
-    |/ldap_kadmind_dn|ldap_service_password_file|ldap_servers/
-    |/ldap_conns_per_server/ in
-    simple_section "dbmodules" keys
+  let subsec_key = /database_name|db_library|disable_last_success/
+    |/disable_lockout|ldap_conns_per_server|ldap_(kdc|kadmind)_dn/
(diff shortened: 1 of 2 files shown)


ptoscano:
ping

ptoscano:
ping -- @lutter ?

lenses/krb5.aug
```diff
@@ -130,10 +131,12 @@ let dbdefaults =
     simple_section "dbdefaults" keys
 
 let dbmodules =
-  let keys = /db_library|ldap_kerberos_container_dn|ldap_kdc_dn/
-    |/ldap_kadmind_dn|ldap_service_password_file|ldap_servers/
-    |/ldap_conns_per_server/ in
-    simple_section "dbmodules" keys
```
raphink: We're losing typecheck on the value keys with this patch, correct?

ptoscano: It depends: the keys previously specified do not seem to be allowed there, see:
http://web.mit.edu/kerberos/krb5-current/doc/admin/conf_files/krb5_conf.html#sections
> Additionally, krb5.conf may include any of the relations described in kdc.conf, but it is not a recommended practice.

So checking the `kdrc.conf` documentation:
http://web.mit.edu/kerberos/krb5-current/doc/admin/conf_files/kdc_conf.html#dbmodules
the only key mentioned as allowed directly for `[dbmodules]` is `db_module_dir`, which is handled in my patch.

... I see what you mean, for the keys within realms. Yes, it can be done, I will amend this soon -- thanks for the notice!

PR #642, opened 2019-08-16
Few improvements to the Mke2fs lens

ptoscano:
- parse more common entries between `[defaults]` and the tags in `[fs_types]`
- fix the type of few entries
- handle the `[options]` stanza

### DIFF
--- a/lenses/mke2fs.aug
+++ b/lenses/mke2fs.aug
@@ -73,20 +73,33 @@ let common_entries_list = ("base_features"|"default_features"|"default_mntopts")
 
 (* View: common_entries_int
     Entries with an integer value *)
-let common_entries_int = ("enable_periodic_fsck"|"flex_bg_size"|"force_undo"
-                         |"inode_ratio"|"inode_size")
+let common_entries_int = ("cluster_size"|"flex_bg_size"|"force_undo"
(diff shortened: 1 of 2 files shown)


lenses/mke2fs.aug
```diff
@@ -34,6 +34,11 @@ let sep = IniFile.sep /=[ \t]*/ "="
 (* View: empty *)
 let empty = IniFile.empty
 
+(* View: boolean value
```
raphink: This needs to be `View: boolean` for the doc to be properly generated

lenses/mke2fs.aug
```diff
@@ -95,24 +130,12 @@ let defaults = IniFile.record defaults_title
  * Group:                 FS_TYPES SECTION
  *************************************************************************)
```
raphink: Is there a reason for reorganizing these entries under common?

ptoscano: You mean all the content of `fs_types_entry`? As I wrote in the commit message, the documentation says that they are allowed both in `[defaults]`, and for each tag in `[fs_types]`. Since `common_entry` represents exactly this, then this is why they were moved.

raphink: Ah right, sorry I missed that!

Issue #652, opened 2019-11-07
Rsyslog.lns does not support multiple actions


Salokyn:
Hi,

I've been confronted to rsyslog files containing this kind of syntax:
```
:msg, startswith, "iptables:" -/var/log/iptables.log
& ~
```
I get the following error:
```
Error in /etc/rsyslog.d/iptables.conf:2.0 (parse_failed)
  Get did not match entire input
  Lens: /usr/share/augeas/lenses/dist/rsyslog.aug:87.10-.43:
```

That should be read as:
```
/files/etc/rsyslog.d/iptables.conf/filter
/files/etc/rsyslog.d/iptables.conf/filter/property = "msg"
/files/etc/rsyslog.d/iptables.conf/filter/operation = "startswith"
/files/etc/rsyslog.d/iptables.conf/filter/value = "iptables: "
/files/etc/rsyslog.d/iptables.conf/filter/action[1]
/files/etc/rsyslog.d/iptables.conf/filter/action[1]/no_sync
/files/etc/rsyslog.d/iptables.conf/filter/action[1]/file = "/var/log/iptables.log"
/files/etc/rsyslog.d/iptables.conf/filter/action[2]
/files/etc/rsyslog.d/iptables.conf/filter/action[2]/discard
```

This lens is too complicated for me to change it myself :sweat_smile:

raphink:
How does that syntax work? Is it `\n& ` or can it be on one line?

Salokyn:
From https://www.rsyslog.com/doc/master/configuration/actions.html#legacy-description

> You can have multiple actions for a single selector (or more precisely a single filter of such a selector line). Each action must be on its own line and the line must start with an ampersand (‘&’) character and have no filters. An example would be
>
> *.=crit :omusrmsg:rger
>& root
>& /var/log/critmsgs
>
> These three lines send critical messages to the user rger and root and also store them in /var/log/critmsgs. Using multiple actions per selector is convenient and also offers a performance benefit. As the filter needs to be evaluated only once, there is less computation required to process the directive compared to the otherwise-equal config directives below:
>
>*.=crit :omusrmsg:rger
> *.=crit root
> *.=crit /var/log/critmsgs

raphink:
OK, then I think https://github.com/hercules-team/augeas/pull/653 should fix it

ptoscano:
Hi @raphink,

in RHEL we have a bug about this for a while now, and I did not find the time to properly tackle it:
[...]

PR #653, opened 2019-11-08
Rsyslog: support multiple actions in filters and selectors

raphink:


ptoscano:
It seems to work fine with the test case I mentioned in #652 -- thanks!

Salokyn:
Works for me :+1:

PR #661, opened 2020-02-11
Mke2fs: support quoted values

ptoscano:
`mke2fs.conf(5)` says that both tag names and values can be quoted using double quotes, in case they contain spaces. Sadly, the current Mke2fs lens does not support that.

In order to support this, introduce an helper `fspath` value to replace `Rx.fspath`, as the latter allows `"` as possible character (which we do not want in quoted strings).

Then change the helper `list_sto` and `entry_sto` functions to allow optionally double quoted values, with no changes in how the tree looks like.

Moreover, try to support also empty values, although only for lists at the moment.

### DIFF
--- a/lenses/mke2fs.aug
+++ b/lenses/mke2fs.aug
@@ -39,6 +39,9 @@ let empty = IniFile.empty
     for booleans, so list all the recognized values *)
 let boolean = ("y"|"yes"|"true"|"t"|"1"|"on"|"n"|"no"|"false"|"nil"|"0"|"off")
 
+(* View: fspath *)
+let fspath = /[^ \t\n"]+/
+
 
 (************************************************************************
  * Group:                 RECORD TYPES
@@ -52,15 +55,14 @@ let entry (kw:regexp) (lns:lens) = Build.key_value_line kw sep lns
 
 (* View: list_sto
     A list of values with given lens *)
-let list_sto (kw:regexp) (lns:lens) = counter "item" .
-                                          entry kw
-                                            (Build.opt_list 
-                                              [lns]
-                                              Sep.comma)
+let list_sto (kw:regexp) (lns:lens) =
+  entry kw (Quote.do_dquote_opt_nil (Build.opt_list [lns] Sep.comma))
 
 (* View: entry_sto
     Store a regexp as entry value *)
-let entry_sto (kw:regexp) (val:regexp) = entry kw (store val)
+let entry_sto (kw:regexp) (val:regexp) =
+  entry kw (Quote.do_dquote_opt_nil (store val))
+  | entry kw (Util.del_str "\"\"")
 
 
 (************************************************************************
@@ -114,7 +116,7 @@ let common_entry   = list_sto common_entries_list (key Rx.word)
 (* View: defaults_entry
     Possible entries under the <defaults> section *)
 let defaults_entry = entry_sto "fs_type" Rx.word
-                   | entry_sto "undo_dir" Rx.fspath
+                   | entry_sto "undo_dir" fspath
                    
 (* View: defaults_title
     Title for the <defaults> section *)
--- a/lenses/tests/test_mke2fs.aug
+++ b/lenses/tests/test_mke2fs.aug
@@ -85,6 +85,40 @@ module Test_mke2fs =
         { "sync_kludge" = "1" } }
 
 
+   let quoted_conf = "[defaults]
+	base_features = \"sparse_super,filetype,resize_inode,dir_index,ext_attr\"
+
+[fs_types]
+	ext4dev = {
+		features = \"has_journal,^extent\"
+		default_mntopts = \"user_xattr\"
+		encoding = \"utf8\"
+		encoding = \"\"
+	}
+"
+
+   test Mke2fs.lns get quoted_conf =
+     { "defaults"
+        { "base_features"
+             { "sparse_super" }
+             { "filetype" }
+             { "resize_inode" }
+             { "dir_index" }
+             { "ext_attr" } }
+        {} }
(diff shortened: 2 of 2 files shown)


ptoscano:
As I mentioned, I wanted to get also empty values for list tags to work.
Sadly, I cannot despite my efforts, and most probably I'm missing something.

What I tried is (on top of this MR):
```diff
diff --git a/lenses/mke2fs.aug b/lenses/mke2fs.aug
index 0df74210..b10e156b 100644
--- a/lenses/mke2fs.aug
+++ b/lenses/mke2fs.aug
@@ -57,6 +57,7 @@ let entry (kw:regexp) (lns:lens) = Build.key_value_line kw sep lns
     A list of values with given lens *)
 let list_sto (kw:regexp) (lns:lens) =
   entry kw (Quote.do_dquote_opt_nil (Build.opt_list [lns] Sep.comma))
+  | entry kw (Util.del_str "\"\"")
 
 (* View: entry_sto
     Store a regexp as entry value *)
diff --git a/lenses/tests/test_mke2fs.aug b/lenses/tests/test_mke2fs.aug
index bcf2fe61..24acac46 100644
--- a/lenses/tests/test_mke2fs.aug
+++ b/lenses/tests/test_mke2fs.aug
@@ -93,6 +93,7 @@ module Test_mke2fs =
                features = \"has_journal,^extent\"
                default_mntopts = \"user_xattr\"
                encoding = \"utf8\"
+               default_mntopts = \"\"
                encoding = \"\"
        }
 "
@@ -115,6 +116,7 @@ module Test_mke2fs =
              { "default_mntopts"
                 { "user_xattr" } }
              { "encoding" = "utf8" }
+             { "default_mntopts" }
              { "encoding" }
              } }
 
```
which results in:
```
SRCDIR/lenses/mke2fs.aug:98.0-111.39:Failed to compile common_entry
SRCDIR/lenses/mke2fs.aug:59.2-60.34:exception: overlapping lenses in tree union.put
    Example matched by both:  { "default_features" }
    First lens: SRCDIR/lenses/build.aug:105.35-.63:
    Second lens: SRCDIR/lenses/build.aug:105.35-.63:
```

Any idea?

ptoscano:
(let's try to invoke mater @lutter for ideas :-) )

ptoscano:
ping?

ptoscano:
ping?

lenses/mke2fs.aug
```diff
@@ -39,6 +39,9 @@ let empty = IniFile.empty
     for booleans, so list all the recognized values *)
 let boolean = ("y"|"yes"|"true"|"t"|"1"|"on"|"n"|"no"|"false"|"nil"|"0"|"off")
 
+(* View: fspath *)
+let fspath = /[^ \t\n"]+/
```
raphink, line 43: I wonder if we should modify `Rx.fspath` to this. That'd be a breaking change though...

raphink:
This automatic quoting thing has always been a bit complex.

If you look at other lenses, `do_dquote_opt_nil` is used for bare entries, while `do_dquote` is used for quoted entries, e.g. in `inifile.aug`:

```augeas
  in let bare = Quote.do_dquote_opt_nil (store (bare_re_noquot . (bare_re* . bare_re_noquot)?))
  in let quoted = Quote.do_dquote (store (no_quot . comment_re+ . no_quot))
  in [ kw . sep . (Sep.opt_space . bare)? . (comment|eol) ]
   | [ kw . sep . Sep.opt_space . quoted . (comment|eol) ]
```


I unfortunately never found a way to deal with both situations in one expression.

lenses/mke2fs.aug
ptoscano: Yeah, IMHO it is too risky to change it for every user of `Rx`.

ptoscano:
> This automatic quoting thing has always been a bit complex.
> 
> If you look at other lenses, `do_dquote_opt_nil` is used for bare entries, while `do_dquote` is used for quoted entries, e.g. in `inifile.aug`:
> 
> ```
>   in let bare = Quote.do_dquote_opt_nil (store (bare_re_noquot . (bare_re* . bare_re_noquot)?))
>   in let quoted = Quote.do_dquote (store (no_quot . comment_re+ . no_quot))
>   in [ kw . sep . (Sep.opt_space . bare)? . (comment|eol) ]
>    | [ kw . sep . Sep.opt_space . quoted . (comment|eol) ]
> ```
> 
> You'll find the same types of constructs in `mysql.aug` and `toml.aug`.
> 
> I unfortunately never found a way to deal with both situations in one expression.

I saw those bits (the `Quote` lens as a whole), and I tried to apply it in the past, apparently without success. Also the challenge here is that the content of the optionally quoted value in some cases are lists of comma-separated items.

raphink (approved):
LGTM

rwmjones:
I don't think this fix is correct.  It does not parse:
```default_mntopts = ""```
(whereas the real parser does).

I checked the actual homebrew parsing code in e2fsprogs, and what it does is simply to parse a string optionally enclosed in quotes.  Different code is then responsible for splitting that string into a list.

So `""` (or nothing at all) should parse as an empty list, `abc` or `"abc"` would both be single element lists, `"abc,def"` would be a two element list, etc.

However my attempts to translate this into lens code has not gone well so far.

ptoscano:
> I don't think this fix is correct. It does not parse: `default_mntopts = ""` (whereas the real parser does).

The fix is correct, for what it does. It simply not complete, and if you read the description and the discussion that follows (which most likely you haven't) you will see that I explicitly mentioned:

| Moreover, try to support also empty values, although only for lists at the moment.

asking for extra help on this.

PR #663, opened 2020-02-13
Krb5: handle [plugins] subsection

ptoscano:


ptoscano:
ping?

jcpunk:
A quick test on the Fermi config has this working for some of our plugin plans.

lenses/tests/test_krb5.aug
```diff
@@ -1071,3 +1071,37 @@ test Krb5.lns get dbmodules_test =
...
+test Krb5.lns get plugins_test =
+  { "plugins"
+    { "clpreauth"
+      { "module" = "mypreauth:/path/to/mypreauth.so" }
+    }
+    { "ccselect"
+      { "disable" = "k5identity" }
+    }
+    { "pwqual"
+      { "module" = "mymodule:/path/to/mymodule.so" }
```
jcpunk, line 1101: Would this be easier to work with as a deeper tree where `module` is a `seq` with `mymodule` = `/path/to/mymodule.so` so folks who need to easily switch from `/usr/lib` to `/usr/lib64` for a specific module can target more directly?

lenses/tests/test_krb5.aug
```diff
@@ -1071,3 +1071,37 @@ test Krb5.lns get dbmodules_test =
...
     { "db_module_dir" = "/some/path" }
   }
+
+(* [plugins] test *)
+let plugins_test = "[plugins]
+    clpreauth = {
+        module = mypreauth:/path/to/mypreauth.so
+    }
+    ccselect = {
+        disable = k5identity
```
jcpunk, line 1081: `disable` and `enable_only` might need to be a `seq` as it can be specified multiple times https://web.mit.edu/kerberos/krb5-1.18/doc/admin/host_config.html

lenses/tests/test_krb5.aug
ptoscano: I'm not sure how `seq` would help in the case you mentioned, since you can use the following expressions to find your node:
- `/files/etc/krb5.conf/plugins/*/module[. =~ regexp(".*/mymodule2.*")]`
- `/files/etc/krb5.conf/plugins/*/module[. =~ regexp("mymodule:.*")]`

ptoscano:
@jcpunk (replying here about both your notes wrt `seq`)
I thought about using `seq` intially, then I had the following thoughts:
- other parts of krb5.conf can have repeated keys, and there is no `seq` used; thus using `seq` here would make it a bit incoherent with the existing layout of the file
- using `seq` would mean having "normal" keys and `seq` keys mixing together, which IMHO is a bit ugly
- if any existing key will be allowed to be repeated in the future, then it will not have a `seq` subtree (and doing so would mean breaking the compatibility of the generated tree, which I don't think it is a good idea)

jcpunk:
Good call, I think all that makes sense.

raphink (approved):
LGTM

lenses/tests/test_krb5.aug
raphink: I'd avoid `seq` as much as possible when it's not necessary to use it. Imo `seq` is only useful when several consecutive nodes have the same label and you want to make sure they don't exchange indent/spaces when one of them is removed.

PR #666, opened 2020-03-12
lib: remove extra @LIBS@ from pkg-config file

ptoscano:
At the moment it is empty, so probably it does not exist. Remove it to avoid adding spurious content to the pkg-config file in case that variable will get a value in the future.


ptoscano:
ping?

lutter:
Sorry for the extremely tardy reply. Looks good!

PR #670, opened 2020-04-28
set error before return in parse_int

ltx2018:
if used==0 in parse_int , parse->error should be set before return


ltx2018:
ping ?

ltx2018:
fix #671 
ping @raphink

PR #675, opened 2020-05-28
lense: cockpit.conf

jcpunk:
Now with working unit tests

### DIFF
--- a/lenses/tests/test_cockpit.aug
+++ b/lenses/tests/test_cockpit.aug
@@ -1,0 +1,94 @@
+module Test_cockpit =
+(************************************************************************
+ * Basic Parse
+ *************************************************************************)
+
+   let conf = "
+[WebService]
+Origins = https://somedomain1.com https://somedomain2.com:9090
+
+MaxStartups= 10:30:60
+[Log]
+Fatal =criticals warnings
+
+[Session]
+
+IdleTimeout=15
+"
+
+   test Cockpit.lns get conf =
+  {  }
+  { "WebService"
+    { "Origins" = "https://somedomain1.com" }
+    { "Origins" = "https://somedomain2.com:9090" }
+    {  }
+    { "MaxStartups" = "10:30:60" }
+  }
+  { "Log"
+    { "Fatal" = "criticals" }
+    { "Fatal" = "warnings" }
+    {  }
+  }
+  { "Session"
+    {  }
+    { "IdleTimeout" = "15" }
+  }
+
+(************************************************************************
+ * Insert simple
+ *************************************************************************)
+    test Cockpit.lns put conf after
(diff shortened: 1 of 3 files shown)


lenses/tests/test_cockpit.aug
```diff
@@ -0,0 +1,94 @@
...
+"
+
+
+(************************************************************************
+ * Insert Origin makes sense
+ *************************************************************************)
+
+    test Cockpit.lns put conf after
+       insa "Origins" "WebService/Origins[last()]";
+       set "WebService/Origins[last()]" "https://thirdhost.com:8080"
```
raphink, line 63: Not sure `insa` is necessary here. You could simple use:

```suggestion
       set "WebService/Origins[.='https://thirdhost.com:8080']" "https://thirdhost.com:8080"
```

I think

raphink: Ah no, my bad, it's because of `MaxStartups`.

Don't you want to use a `seq` inside the `Origins` entry instead? That would be easier to manipulate:

```
{ "WebService"
  { "Origins"
    { "1" = "https://somedomain1.com" }
    { "2" = "https://somedomain2.com:9090" } }
  { "MaxStartups" = "10:30:60" } }
```

jcpunk: I wasn't sure what the preferred approach was on that front.

If you think a `seq` would be cleaner I can try to make that work... I'll have to re-read the doc on that operator.

jcpunk: checking back in....

jcpunk: Ping

georgehansper: Do you want to keep this lens as-is, or modify it to use a `seq` ?

I can't see an obvious advantage to one approach over the other

jcpunk: I'd probably need a hand converting to a `seq`.  Mostly due to it being a bit since I've played with augeas.

jcpunk:
I'd probably need a hand converting to a `seq`.  Mostly due to it being a bit since I've played with augeas. If that is still something folks would like to see in this lense.

lenses/tests/test_cockpit.aug
georgehansper: Sorry I could not be of help here.
I still think it's an abitrary change, and most lenses that I can find seem to use the same approach ie.
```
/files/etc/cockpit/cockpit.conf/WebService/Origins[1] = "https://somedomain1.com"
/files/etc/cockpit/cockpit.conf/WebService/Origins[2] = "https://somedomain2.com:9090"
```

georgehansper (approved):
Looks good to me

PR #677, opened 2020-06-04
Add additional unit tests for yum lense

jcpunk:
When writing a new lense using logic from this lense I uncovered unexpected behaviour with whitespace.

The tests added here expose this oddity.

### DIFF
--- a/lenses/tests/test_yum.aug
+++ b/lenses/tests/test_yum.aug
@@ -75,6 +75,12 @@ gpgcheck=1
 gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-remi
 "
 
+  let epel_repo = "[epel]
+name=Extra Packages for Enterprise Linux 6 - $basearch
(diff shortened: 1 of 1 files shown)


raphink:
Thanks. I usually add a `needs-tests` label to PRs without tests, but this one has tests and no code...

jcpunk:
My primary worry is that I'm doing the testing wrong.... if I'm not, this is a real bug that I'm not sure how to fix.

jcpunk:
After a night's sleep, I realized what was wrong with my tests.  In theory I'm now running the tests right and should have more coverage than before.

jcpunk:
rebased off master

jcpunk:
rebased off master

georgehansper:
This change does not fail any tests - were you expecting it to fail on the yum lens ?

jcpunk:
Nope i was just adding some coverage to the yum lense and providing a
syntax cheat sheet for when i do this sort of thing.

PR #685, opened 2020-07-28
Add new command 'update' which changes a single value, but never creates a new node

georgehansper:
This Pull Request adds a new command 'update' to augtool, and a corresponding new function call 'aug_update' to the API.

'update' does what 'set' does, except that it expects the path to already exist.

'update' will never create new nodes or path-components. while 'set' will create a new node and any required path-components required by the new node

```
   update /files/etc/hosts/1/ipaddr 127.0.0.2
```

is equivalent to

```
   set /files/etc/hosts/1/ipaddr 127.0.0.2
```

(assuming /etc/hosts contains at least one valid entry).

```
   update /files/etc/hosts/00/ipaddr 127.0.0.2
```

will make no changes, while

```
   set /files/etc/hosts/00/ipaddr 127.0.0.2
```

will create the new node 'ipaddr' and it's parent '00' and assign the value.

The API call aug_update() returns 

```
   1  on success
   0  if no existing node matches PATH (this is an error)
  -1  on error, or if multiple nodes matched PATH (this is an error)
```

This is in alignment with aug_get() aug_label() and aug_setm(), which all return 0 if there are no matching nodes
This is notably different from aug_set(), which returns 0 on success.

'update' sets an error if there is no matching node.
'setm' could be used in the case where no matching node is not considered an error.

Please let me know if any of these behaviours should be adjusted.


lutter:
This looks great - the only thing that I think should be adjusted is how to deal with the situation where no node matches `path`. If that is considered an error, I think we should return `-1` to indicate to the caller that they need to inspect `aug_error`. If it's not an error, we should return `0`, but not flag it as an error. I would lean towards the latter, i.e., `aug_update` means "update that node if it exists" and the caller can check the return value if anything changed or not.

In the API, I have been (trying to) stick to the convention that a negative return value means 'inspect `aug_error()`' and a non-negative return value means 'the return value tells you everything you need to know about what happened'.

georgehansper:
Thanks for your feedback.

I do like the idea that the return code from aug_update() makes it easy to distinguish between the 'no matching node' case, and 'too many nodes matched' case.

The closest operations to `update` are `get` and `label`, in that they also require exactly one node to match in order to perform their stated operation.

I notice now that both `get` and `label` do not treat 'no matching nodes' as an 'error', even though they do not return a value.
In `augtool`, they print a visual indication that no nodes matched, eg
```
augtool> get /files/yellow_brick_road
/files/yellow_brick_road (o)
augtool> 
```

I can suggest 4 possibilities:
1. (The current PR) no-match is an error, and returns 0 to the API call `aug_update()`
2. (Your suggestion) no-match is an error, and returns -1 to the API call `aug_update()`
3.  no-match is not an error, returns 0 to the API call and silently does nothing in `augtool`
4. no-match is not an error, returns 0 to the API call, and `augtool` prints a message similar to `get` and `label` when there is no match

I'm thinking (4) might be the closest to the existing `get` and `label` operations, while still utilising the return-value from aug_update() to distinguish between outcomes.

Your thoughts ?

lutter:
Yes, I think (4) would be the best option.

georgehansper:
I've updated this pull-request, so that the `update` command and `aug_update()` API call no longer classify "no matching node" as an error. `aug_update()` returns 0 (as before), but no error is stored.

augtool gives the following message, similar to the `get` and `label` commands:

```
augtool> update /files/etc/hosts/99/ipaddr 127.0.0.1
/files/etc/hosts/99/ipaddr (o)
augtool> print /files/etc/hosts/99
augtool> 
```

georgehansper:
This PR was intended to complement #687

PR #687 has been superseded by PR #706 so the usefulness of this PR is diminished.

This PR may be re-opened in the future, if someone considers it useful.

georgehansper:
Reopenning for testing travis-ci.com

Issue #561, opened 2018-06-02
Append new line if not exist using Simplelines.lns


nicutor:
Hi,

I have a file in this format:

```
# comment sample 1
# comment sample 2
line 1 value # comment line 1
line 2 value # comment line 2
Include /some/file
Include /some/other/file
```

I am trying to append a new line at the end of file if that line doesn't already exist using Simplelines and I don't know what I am missing.

Can you please give me an example of how to do this? I am using the right lens?

Thank you!

MatthewHannigan:
Best to show what you've tried and what you got.

nicutor:
Hi @MatthewHannigan,

I found this tutorial https://github.com/hercules-team/augeas/wiki/Adding-nodes-to-the-tree
and i was able to add a new line using this:

```
# augtool -At "Simplelines.lns incl /root/tmp"
augtool> print /files/root/tmp
/files/root/tmp
/files/root/tmp/#comment[1] = "comment sample 1"
/files/root/tmp/#comment[2] = "comment sample 2"
/files/root/tmp/1 = "line 1 value # comment line 1"
/files/root/tmp/2 = "line 2 value # comment line 2"
augtool> ins 01 before /files/root/tmp/1
augtool> set /files/root/tmp/01 'Include /some/file'
augtool> print /files/root/tmp
/files/root/tmp
/files/root/tmp/#comment[1] = "comment sample 1"
/files/root/tmp/#comment[2] = "comment sample 2"
/files/root/tmp/01 = "Include /some/file"
/files/root/tmp/1 = "line 1 value # comment line 1"
/files/root/tmp/2 = "line 2 value # comment line 2"
augtool> save
Saved 1 file(s)
augtool> quit

# cat /root/tmp
# comment sample 1
# comment sample 2
Include /some/file
line 1 value # comment line 1
line 2 value # comment line 2
```

Now I am trying to use the 'match' method to find if the line is already there:
```
augtool> match /files/root/tmp/*[. = 'Include']
  (no matches)
augtool> match /files/root/tmp/*[. =~ 'Include']
error: Invalid path expression
error: type error
/files/root/tmp/*[. =~ 'Include']|=|
augtool>
```

What I am trying to do is, to add some 'Include /path/to/some/files' lines on the file, but only if the line is not already there, no matter at what position. 

The match function will return a code so I can use it on this chef resource:
https://supermarket.chef.io/cookbooks/augeas
For example:
```
augeas 'add_included_files' do
[...]

Issue #68, opened 2013-12-30
make idempotent changes easy with augtool


rhoerbe:
as discussed in (4) in http://www.redhat.com/archives/augeas-devel/2010-January/msg00100.html:

Using augtool from scripts really requires a significant wrapper to handle the frequent use case of idempotently adding a line:
- add a node (subtree) if it doesn't exist
- do nothing if it exists with the same values as set
- if an exiting node exists with deviating values complain and abort or, if requested, set the values to the desired state.
- handle errors reliably

.. to really support idempotent changes, augtool will need to support an if statement; in general, you will make multiple changes if something in the tree isn't what you expect (e.g., if there's no entry in /etc/hosts for 192.168.0.1, you'll want to add several nodes)

I am thinking of a syntax like
        if PATHEXPR then
          .. statements (one per line) ..
        else
          .. statements ..
        end

As a result, this syntax should allow changes like:
- adding an alias to /etc/hosts for a specific ip-address only if it does not exist for this node
- if the alias in the example above exists for another node, the alias can be either deleted for the other node, or an error be thrown (exit > 0)
- insert following  line into pg_hba.conf if it does not exist
  host    all             postgres        127.0.0.1/32            ident
- choose either to throw an error or to update an existing line that matches
  host    all             postgres        127.0.0.1/32 

I would be willing to offer some $ on bountysource - someone interested to join?

ioquatix:
Does this work yet?

lutter:
No, none of this was ever implemented. It's a good amount of work, and I am a little hesitant to add this since it creates a completely new language and I am a concerned that this language will become more complex over time.

ioquatix:
I see. What about exposing augtool as a library and then implementing, say, something in Ruby or Python on top?

lutter:
That's a lot easier, and there are a number of augeas bindings for various languages ;) See [the download page](http://augeas.net/download.html) for a list of bindings

ioquatix:
If that's the outcome, I think you can close this issue with wont fix :)

lutter:
The simplest way I can think of to make idempotent changes to the tree is using the following approach

- set your desired state up somewhere outside of `/files`, say at `/scratch`
- check if the target node exists already; if it does move `/scratch` there
- if not move `/scratch` to its (new) place

This is idempotent since Augeas will not modify the underlying file if what was build up in `/scratch` was already in the tree. The last part, for example in Ruby, would look like
```ruby
def update(aug, scratch, target, create)
[...]

PR #687, opened 2020-07-31
Add augtool option --createifnomatch to make 'set' create a new node if a path-filter fails to match

georgehansper:
This Pull Request adds the following to augeas:

* a new augtool command-line option `--createifnomatch` or `-C`
  If specified, this changes the operation of the `set` command
  to create a new, numbered node if the path-filter fails to match
  The new node is named according the highest numbered node already
  in use, plus one.
* a new environment variable AUGEAS_CREATEIFNOMATCH that does the same as
  the command-line option `--createifnomatch` if it is set to a non-zero value
* it creates a node to the `/augeas` tree `/augeas/pathx/createifnomatch`
  if either of the above are used

The node `/augeas/pathx/createifnomatch` does not modify the `set` behaviour on-the-fly.
It is only an indicator that this option (or environment variable) has been used.

Without `--createifnomatch` a set command will fail if the `*[expr]` matches nothing:

```
root@localhost > augtool
augtool> print /files/etc/hosts/*/ipaddr
/files/etc/hosts/1/ipaddr = "127.0.0.1"
/files/etc/hosts/2/ipaddr = "::1"
augtool> set /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr "8.8.8.8"
Setting /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr failed
augtool> 
```

With `--createifnomatch` a set command will append a new node if the `*[expr]` matches nothing:

```
root@localhost > augtool --createifnomatch
augtool> print /files/etc/hosts/*/ipaddr
/files/etc/hosts/1/ipaddr = "127.0.0.1"
/files/etc/hosts/2/ipaddr = "::1"
augtool> set /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr "8.8.8.8"
augtool> print /files/etc/hosts/*/ipaddr
/files/etc/hosts/1/ipaddr = "127.0.0.1"
/files/etc/hosts/2/ipaddr = "::1"
/files/etc/hosts/3/ipaddr = "8.8.8.8"
augtool> 
```

This Pull Request allows `set` commands to be crafted so as to be idempotent, for the specific case of
sequential nodes, such as entries in /etc/hosts, /etc/fstab, and /etc/aliases

For example, adding a new hosts entry, using a idempotent `set` command:

```
root@localhost > augtool --createifnomatch
augtool> print /files/etc/hosts/*/ipaddr
/files/etc/hosts/1/ipaddr = "127.0.0.1"
/files/etc/hosts/2/ipaddr = "::1"
augtool> set /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr "8.8.8.8"
augtool> set /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr "8.8.8.8"
augtool> set /files/etc/hosts/*[ipaddr="8.8.8.8"]/ipaddr "8.8.8.8"
augtool> print /files/etc/hosts/*/ipaddr
/files/etc/hosts/1/ipaddr = "127.0.0.1"
/files/etc/hosts/2/ipaddr = "::1"
/files/etc/hosts/3/ipaddr = "8.8.8.8"
augtool> 
```

For example, an idempotent change to /etc/aliases would look like this:

```
root@localhost > augtool --createifnomatch
augtool> print /files/etc/aliases/*[name="postgres"]
augtool> set /files/etc/aliases/*[name="postgres"]/name "postgres"
augtool> set /files/etc/aliases/*[name="postgres"]/value "dba@example.com"
augtool> print /files/etc/aliases/*[name="postgres"]
/files/etc/aliases/76
/files/etc/aliases/76/name = "postgres"
/files/etc/aliases/76/value = "dba@example.com"
augtool> set /files/etc/aliases/*[name="postgres"]/name "postgres"
augtool> set /files/etc/aliases/*[name="postgres"]/value "dba@example.com"
augtool> print /files/etc/aliases/*[name="postgres"]
/files/etc/aliases/76
/files/etc/aliases/76/name = "postgres"
/files/etc/aliases/76/value = "dba@example.com"
augtool> 
```

If there is no existing node matching `/files/etc/aliases/*[name="postgres"]` a new node is created by the command:

```
set /files/etc/aliases/*[name="postgres"]/name "postgres"
```

If `--createifnomatch` is not specified, the above `set` command would fail.

This is the existing behaviour, and the default behaviour if `--createifnomatch` is not specified.

The issue of making idempotent changes from augeas is discussed in issue #68 and issue #561 

This Pull Request makes idempotency feasible using a one-line set command, without introducing if-then style conditional command execution.


georgehansper:
Closing in favour of PR #706

PR #689, opened 2020-08-06
src/fa.c: fixed error in regexp grammar in comment

safinaskar:
Fixed possible error.

Also: I'm trying to build my project (my reversible analog of flex) on top of libfa. I can tell more if you want


safinaskar:
Also, grammar currently allows `[a]`, but forbids `[ab]`. Please, review grammar. I want to know true grammar, because I am going to actually use libfa API

lutter:
Yes, you are right - the grammar was wrong. It should also have
```
 char_class: '[' char_exp+ ']'
            | '[' '^' char_exp+ ']'
```

I just fixed that, too.

PR #690, opened 2020-08-12
Add CLI command 'preview' and API 'aug_preview' to preview file contents

georgehansper:
This Pull Requests adds a new command `preview` to augtool, and a new function `aug_preview()` to the API.

The `preview` command accepts a path-name as an argument.

For the file corresponding to the path-name, the contents of the file are written to stdout, with the current
nodes and values from the `/files` tree applied.

This corresponds to the contents of the file that would be written by a subsequent `save` operation.

The path-expression must evaluate to a single node only, and the source-file is determined from this node.

The API call `aug_preview()` places the resulting file-contents into `char *OUT`, allocating the required memory.

        int aug_preview(augeas *aug, const char *path, char **out);

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -80,7 +80,8 @@ static const char *const errcodes[] = {
     "Failed to execute command",                        /* AUG_ECMDRUN */
     "Invalid argument in function call",                /* AUG_EBADARG */
     "Invalid label",                                    /* AUG_ELABEL */
-    "Cannot copy node into its descendant"              /* AUG_ECPDESC */
+    "Cannot copy node into its descendant",             /* AUG_ECPDESC */
+    "Cannot access file"                                /* AUG_EFILEACCESS */
 };
 
 static void tree_mark_dirty(struct tree *tree) {
@@ -1972,6 +1973,7 @@ int aug_preview(struct augeas *aug, const char *path, char **out) {
     struct tree *tree = NULL;
     struct pathx *p;
     int r;
+    int result=-1;
     char *lens_path = NULL;
     char *lens_name = NULL;
     char *file_path = NULL;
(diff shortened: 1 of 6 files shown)


raphink:
That looks very interesting!

At the moment, this feature is achieved in various tools by leveraging the `new` mode and diffing the files. Doing it in memory would be much nicer.

georgehansper:
The preview function has to read the original file (again) in order to apply the transform to it, so it has to do a 2nd read of the original file.

It does save having to write the file out and having to read the resulting file back off the disk (by whatever is wants to do the diff), so it saves IO operations.

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+
+    if ( file_path == NULL ) {
+        file_path = strdup(path);
+        ERR_NOMEM(file_path == NULL, aug);
+    } else {
+        tree = tree_find(aug, file_path);
+    }
+
+    r = xasprintf(&lens_path, "%s%s/%s", AUGEAS_META_TREE, file_path, s_lens);
+    if ( r>= 0 ) {
```
lutter, line 2004: This is super nitpicky, but I generally write `if` statements with now spaces after the parentheses, but spaces around the operator, i.e., `if (r >= 0) { ..` - I wish there was some easy-to-use code formatter that we could just run to make sure these things are consistent. In any event, I would prefer to have these things unfirom as much as possible, but it's a preference, not a "I would never merge a PR formatted like this" ;)

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+        r = aug_get(aug,lens_path,(const char **) &lens_name);
```
lutter, line 2005: You don't need the assignment to `r` in what follows - better to leave that out so the reader doesn't wonder why it's assigned to.

src/augeas.c
lutter: Instead of the `if` statement, it would be better to write
```
ERR_NOMEM(lens_path == NULL, aug)
```

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+        ERR_BAIL(aug);
+    }
+
+    ERR_THROW(lens_name == NULL, aug, AUG_ENOLENS, "No lens found for path %s", path);
+
+    if ( STREQLEN(file_path, AUGEAS_FILES_TREE, strlen(AUGEAS_FILES_TREE) ) ) {
```
lutter, line 2011: This `if` does not seem necessary - we already constructed `file_path` to have the 'right' content, so we should be able to perform the body of the `if` unconditionally.

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+        xasprintf(&source_filename, "%s%s",aug->root, file_path + strlen(AUGEAS_FILES_TREE));
+        ERR_NOMEM(source_filename == NULL, aug);
+        source_text = xread_file(source_filename);
+    }
+    if ( source_text == NULL ) {
```
lutter, line 2016: I think a `NULL` here should be reported as an error to the caller - they told us to read a file and we couldn't.

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+    if (r < 0)
+        goto error;
+
+
+    free(p);
+    free(file_path);
+    free(lens_path);
+    free(source_filename);
+    free(source_text);
+    api_exit(aug);
```
lutter, line 2030: Since it's easy to only fix up all the `free` calls in the good path or the error path when this code changes in the future, I usually write functions like this as
```
int something(..) {
  int result = -1;
  .. do things ..
  /* everything worked out */
 result = 0;
error:
  /* free all the things */
 return result;
}
```

tests/test-api.c
```diff
@@ -851,6 +851,45 @@ static void testAugSource(CuTest *tc) {
...
+    char *hosts_txt = NULL;
+    int readsz = 0;
+
+    /* Read the original contents of the etc/hosts file */
+    if (asprintf(&etc_hosts_fn,"%s/etc/hosts",root) >=0 ) {
+        hosts_fp = fopen(etc_hosts_fn,"r");
+        if ( hosts_fp ) {
+            hosts_txt = calloc(sizeof(char),4096);
+            if ( hosts_txt ) {
+							  readsz = fread(hosts_txt,sizeof(char),4096,hosts_fp);
```
lutter, line 869: Looks like there's some errant whitespace

src/augeas.c
```diff
@@ -1968,6 +1968,77 @@ int aug_source(const augeas *aug, const char *path, char **file_path) {
...
+
+    tree = pathx_first(p);
+    ERR_BAIL(aug);
+    ERR_THROW(tree == NULL, aug, AUG_ENOMATCH, "No node matching %s", path);
+
+    file_path = tree_source(aug, tree);
+
+    ERR_THROW(file_path == NULL, aug, AUG_EBADARG, "Path %s is not associated with a file", path);
+
+    if ( file_path == NULL ) {
```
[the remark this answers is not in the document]
georgehansper: `if` removed

src/augeas.c
georgehansper: Replaced `if (r >= 0)` with `ERR_NOMEM` macro

src/augeas.c
georgehansper: Removed unused assignment

src/augeas.c
georgehansper: Removed `if ( STREQLEN`

src/augeas.c
georgehansper: I think that's a better structure. Updated.

tests/test-api.c
georgehansper: Fixed

lutter (approved):

Issue #632, opened 2019-06-03
Ssh: cannot parse Match options


ptoscano:
The Ssh lens cannot parse the `Match` option.
A simple example is in a ssh_config.d snipped shipped in Fedora (30), `/etc/ssh/ssh_config.d/05-redhat.conf`:
```
# The options here are in the "Match final block" to be applied as the last
# options and could be potentially overwritten by the user configuration
Match final all
        # Follow system-wide Crypto Policy, if defined:
        Include /etc/crypto-policies/back-ends/openssh.config

        GSSAPIAuthentication yes

# If this option is set to yes then remote X11 clients will have full access
# to the original X11 display. As virtually no X11 client supports the untrusted
# mode correctly we set this to yes.
        ForwardX11Trusted yes

# Send locale-related environment variables
        SendEnv LANG LC_CTYPE LC_NUMERIC LC_TIME LC_COLLATE LC_MONETARY LC_MESSAGES
        SendEnv LC_PAPER LC_NAME LC_ADDRESS LC_TELEPHONE LC_MEASUREMENT
        SendEnv LC_IDENTIFICATION LC_ALL LANGUAGE
        SendEnv XMODIFIERS

# Uncomment this if you want to use .local domain
# Host *.local
```
See also the documentation for it: https://man.openbsd.org/ssh_config#Match.

The Sshd (for `sshd_config`) lens already handles `Match` options, although it seems they are slightly different than what allowed in `ssh_config`.

raphink:
Which version of Augeas are you using?

ptoscano:
> Which version of Augeas are you using?

It does not matter, since this happens with current git master.

raphink:
I can parse your example fine with current git master.

ptoscano:
> I can parse your example fine with current git master.
> 
> ```
> $ augtool -I lenses/ -At "Sshd.lns incl /tmp/sshd_config" get /files/tmp/sshd_config/Match/Condition/final
> /files/tmp/sshd_config/Match/Condition/final = all
> ```

Yes, the Sshd lens handles it, but not the Ssh lens.

raphink:
Ah, I was not aware that `ssh_config`  supported Match groups!

granquet:
I would be interested in getting this issue fixed.
I toyed with the idea on top of release-1.8.0

I'm discovering augeas, it seems to parse and generate proper ssh_config files.
heavily inspired by sshd.aug

```diff
[...]

PR #695, opened 2020-10-06
Ssh: add Match keyword support

granquet:
Signed-off-by: Guillaume Ranquet <guillaume-externe.ranquet@edf.fr>

Proposal for a fix regarding https://github.com/hercules-team/augeas/issues/632

### DIFF
--- a/lenses/ssh.aug
+++ b/lenses/ssh.aug
@@ -92,7 +92,7 @@ module Ssh =
                         | rekey_limit
 
     let key_re = /[A-Za-z0-9]+/
-               - /SendEnv|Host|ProxyCommand|RemoteForward|LocalForward|MACs|Ciphers|(HostKey|Kex)Algorithms|PubkeyAcceptedKeyTypes|GlobalKnownHostsFile|RekeyLimit/i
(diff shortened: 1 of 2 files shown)


lenses/ssh.aug
```diff
@@ -105,11 +105,26 @@ module Ssh =
...
+     [ label "Condition" . condition_entry+ . eol ]
+
+   let match_entry = entry
+
+   let match =
+     [ key /Match/i . match_cond
+        . [ label "Settings" .  match_entry+ ]
+     ]
+
+
```
raphink, line 122: Or maybe just `let match = Sshd.match` so we reuse the logic from the other lens. What do you think @lutter?

granquet: It doesn't seem to work... the definition of eol in ssh.aug and sshd.aug are different... an induce a change in condition_entry ( [^ \t\r\n] vs [^ \t\n])
not sure how to solve that as I'm just discovering augeas... aside from trying to rewrite ssh.aug and sshd.aug.

raphink: Ok then. Thanks for trying it!

raphink (approved):
Lgtm

Issue #697, opened 2020-10-13
New chrony.conf options unsupported by chrony lens (sourcedir, ntsdumpdir)


flo-renaud:
With Fedora33, chrony-4.0-0.9.pre4.fc33.x86_64 is shipped and provides new options with default values in /etc/chrony.conf that render the lens unusable.

When the default chrony.conf is installed, it contains
```
sourcedir /run/chrony-dhcp
ntsdumpdir /var/lib/chrony
```
and those 2 options break augeas. In order to reproduce, use the following python script:
```
from augeas import Augeas
import os

aug = Augeas(flags=Augeas.NO_LOAD | Augeas.NO_MODL_AUTOLOAD)
module = "chrony"
chrony_conf = os.path.abspath("/etc/chrony.conf")
aug.transform(module, chrony_conf)  # loads chrony lens file
aug.load()  # loads augeas tree

path = '/files{path}'.format(path=chrony_conf)

# remove possible conflicting configuration of servers
aug.remove('{}/server'.format(path))
aug.remove('{}/pool'.format(path))
aug.remove('{}/peer'.format(path))

aug.set('{}/pool[last()+1]'.format(path), "1.pool.ntp.org")
aug.set('{}/pool[last()]/iburst'.format(path), None)

aug.save()
```

The script fails to save chrony.conf with a new pool:
```
# python3 /tmp/test_augeas.py
Traceback (most recent call last):
  File "/tmp/test_augeas.py", line 20, in <module>
    aug.save()
  File "/usr/lib/python3.9/site-packages/augeas.py", line 488, in save
    raise IOError("Unable to save to file!")
OSError: Unable to save to file!
```

If /etc/chrony.conf is manually edited in order to remove the options `sourcedir` and `ntsdumpdir` prior to using augeas, then the script succeeds.

Reproduced with augeas-libs-1.12.0-4.fc33.x86_64 + python3-augeas-0.5.0-22.fc33.noarch

raphink:
Hello,

These options were added in https://github.com/hercules-team/augeas/pull/698. You can pull the new lens from the master branch for now, until a new version is released.

PR #698, opened 2020-10-15
Chrony: add new options

mlichvar:
Add new directives and options supported in chrony since 3.3 up to 4.0.

The ntsport option is unsupported as it conflicts with the nts flag. I wasn't able to figure out how to resolve the conflict. The regex engine doesn't seem to support lookahead.

This should fix issue #697.

### DIFF
--- a/lenses/chrony.aug
+++ b/lenses/chrony.aug
@@ -79,6 +79,7 @@ module Chrony =
          Server/Peer/Pool options with values
     *)
     let cmd_options = "asymmetry"
+                    | "filter"
                     | "key"
                     | /maxdelay((dev)?ratio)?/
                     | /(min|max)poll/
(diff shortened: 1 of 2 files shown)


lenses/chrony.aug
```diff
@@ -254,6 +263,13 @@ module Chrony =
                       . ( space . [ label "port" . store integer ] )?
                       . eol ]
 
+    (* View: bcast
+         confdir and sourcedir have specific syntax
+    *)
+    let dir_list = [ Util.indent . key /(conf|source)dir/
+                      . ( space . [ key no_space ] )+
```
raphink, line 270: This is not the most common way of storing a list of values in Augeas (which would be to use either multiple nodes with the same label, or a `seq`). Is there a reason for doing it this way?

mlichvar: No good reason. I just didn't know the conventions. It should be fixed now. Thanks.

raphink (approved):

PR #706, opened 2020-11-16
Add new axis 'seq' to allow /path/seq::*[expr] to match and create numeric nodes, as idempotent alternative to /path/*[expr]

georgehansper:
This Pull Request adds a wildcard matching symbol `+` which can be used in place of `*`
within path-expressions.

The '+' symbol differs from the '*' wildcard in that the '+' only matches labels which consist soley of decimal digits.

eg:

Wildcard '*' matching
```
augtool> match /files/etc/hosts/*
/files/etc/hosts/#comment[1] = This is the /etc/hosts file
/files/etc/hosts/#comment[2] = Do not delete the localhost entries - many applications rely on them
/files/etc/hosts/1 = (none)
/files/etc/hosts/2 = (none)
augtool> 
```

Plus '+' matching
```
augtool> match /files/etc/hosts/+
/files/etc/hosts/1 = (none)
/files/etc/hosts/2 = (none)
augtool> 
```

This pull-request also adds the feature of idempotency, when '+' matching is used

To create a new, numbered label:

```
augtool> match /files/etc/hosts/+/ipaddr
/files/etc/hosts/1/ipaddr = 127.0.0.1
/files/etc/hosts/2/ipaddr = ::1
augtool> set /files/etc/hosts/+[ipaddr='192.0.2.0']/ipaddr 192.0.2.0
augtool> set /files/etc/hosts/+[ipaddr='192.0.2.0']/canonical test101
augtool> match /files/etc/hosts/+/ipaddr
/files/etc/hosts/1/ipaddr = 127.0.0.1
/files/etc/hosts/2/ipaddr = ::1
/files/etc/hosts/3/ipaddr = 192.0.2.0
augtool> 
```

The above set commands are idempotent, when using the '+' matching symbol:

```
augtool> set /files/etc/hosts/+[ipaddr='192.0.2.0']/ipaddr 192.0.2.0
augtool> set /files/etc/hosts/+[ipaddr='192.0.2.0']/canonical test101
augtool> match /files/etc/hosts/+/ipaddr
/files/etc/hosts/1/ipaddr = 127.0.0.1
/files/etc/hosts/2/ipaddr = ::1
/files/etc/hosts/3/ipaddr = 192.0.2.0
augtool> 
```


raphink:
I like the idea of having an operator to match integer values only, but I'm wondering if `+` is the best choice. The Augeas query language is initially based off XPath, though it has deviated, but XPath has no specific expression to select integer labels.

With `+`, I'm afraid there might be a confusion with the regexp syntax where `*` would match 0, 1 or more, where as `+` would only match 1 and up. But maybe I'm wrong and there would be no confusion.

As for the C implementation, I'd rather @lutter review it than myself 😁

georgehansper:
The choice of operator is really the big question here.

I understand that Augeas filter-path expressions are loosely based on 
Xpath, but in reality, Augeas path-expressions were never compatible 
with XPath expressions.

The need to match integer labels and treat them as "special" is not just 
a "new feature". It is driven by the seq primitive in Augeas lenses. The 
seq primitive generates labels automatically. It is possible to /match 
/an integer label with the *[ expr ] operators, but if we are in a 'set' 
operation, there is no label to assign to the new node, because '*' does 
not specify a label.

'set' operations are not part of Xpath, but they are part of Augeas.

So the question for me is: what solution suits _Augeas_ best.

Currently, there are 2 kinds of filter-path expressions:

  * label[ expr ]
  * *[ expr ]

One interesting thing about Augeas is that the 'label' is largely abitrary.

If we are to introduce a new operator, it should be something that will 
not conflict with label

Any string could be used (although the code is simple if it's a single 
character). A few single-character possibilities spring to mind:

  * *?*
      o This is already widely recognised as a 'special' character, but
        frequently needs escaping for exactly that reason
  * *+*
      o Already associated with regular-expressions (good), and usually
        does not need escaping
  * *#*
      o Widely used to introduce comments (otherwise probably the best
        solution)
  * *0*
      o Doesn't conflict with anything at the moment. Not very
        suggestive of actually being a pattern rather than a label.

Or we could use a short string, such as:

  * *num*
  * *number*
  * *%d*
  * *\d+*
  * *%i*
  * *seq*

The problem with things like *%d* is that they all suggest a richer 
syntax than currently exists.

I cannot forsee further extensions to the syntax being required, as we 
are just trying to accomodate labels generate by the seq lens primitive

I'm not too endeared to any of the above possibilities, and would 
happily implement _any_ solution that extends the idempotency of Augeas 
to include seq labels.

But for what it's worth, probably the *# *character would make a good 
alternative to the *+*

augtool will ignore full-line comments like this:

# some text

but does not support trailing comments

set /files/etc/hosts/2/ipaddr "::1" # some text

So the *#* character could easily be used to match integers. It has the 
advantage that it would not be confused or conflict with 
regular-expression syntax, or other common practices.

Please let me know what you think.

georgehansper:
After some consideration, I think '#' is a better choice than '+' for matching integers.

It doesn't conflict with anything else as far as I can tell.

lutter:
On the question of how to fit that into the path expression language, it seems cleaner to me to do this either with a function or as another match axis, both are concepts that come from XPath and are currently used in path expressions: you can write stuff like `/some/thing[position() = 3]` (which is the explicit version of `/some/thing[3]`. You can also write somethind like `/some/descendant::thing` which will search the entire tree underneath `/some` for nodes labelled `thing`.

My suggestion therefore would be to enable this functionality (which I like a lot, I think it is very useful) with either the syntax `/some/*[seq()]/thing` or `/some/seq::*/thing`. There might be better names than `seq` for this, but it does mirror how such labels generally come about.

Caveat: I haven't thought at all how hard either of these would be to implement, but I think especially the `*[seq()]` form wouldn't be very hard.

georgehansper:
`some/*[seq()]/thing`

Keep in mind that the key functionality of this PR is not just to filter nodes, but also to allow new sequential nodes to be created by the `set` command, if no matching node exists.

Creating a new node is currently not possible with the existing `*[filter]` expressions, because the `*` does not set a label for the new node.

The `#` match sets a flag saying that 'if no nodes match, choose the next number in the sequence as the new label'

It would be possible to create a filter-function `seq()` that matches only numeric labels, but I seem to recall that passing back the flag mentioned above would require substantial changes to the code, because currently filter-expressions only return a boolean value.

There is no current provision for a filter-function to perform additional actions, so I thought this would not be a good fit to what's there now.

`/some/seq::*/thing`

This is quite simple to implement with relatively minor changes to the current PR (plus some code for accommodating the new `seq` axis).

Functionally, it is almost equivalent to `#`, except that it is possible specify `seq::3` and select only the node with label `3`
(This is not a practical use-case, because `/some/3/thing` can be explicitly specified without referring to the axis.)

Specifying `seq::text` would _always_ append a new sequential node, because `text` is not numeric, and hence a node cannot match both the `seq` criteria and the explicit label `text`.
This could provide an alternative to the label `0` when a new sequential node is desired eg. `seq::new`

I like the axis name `seq::*` because of the tie-in with the lenses.

Personally, I think the `seq::*` syntax is a little less readable than the simpler `#` syntax.

The other reason I prefer `#` over `seq::*` is that I liked the idea that `*[...]` was used purely for updating existing nodes, where as `#[...]` or `label[...]` was explicitly for creating a new node (if missing).

I think it's difficult to stay very close to XPath because (correct me if I'm wrong) XPath is purely for selecting nodes, while Augeas has the capability to create nodes too.

I will update the PR to whatever is the preferred syntax.

georgehansper:
I think the `seq::*` solution is a "good fit" for Augeas, and I've updated the PR to implement it.

moonbaseDelta:
Hello all, I'm trying to use this to create new line for pg_hba.conf but can't figure out how to make it work - #, seq::* and other mensioned are just added to the tree with their names like `/files/tmp/pg_hba.conf/seq::*/type = "host"`.
I built it from `https://github.com/georgehansper/augeas/tree/filter_plus` and from `https://github.com/georgehansper/augeas/tree/filter_seq_axis` yet it seems not to work.

georgehansper:
The `set` command in Augeas can be used to either change an existing value in the tree, or create a new value.

In this regard, it is important that the given path-expression should match exactly one node, or no nodes at all.

To understand what `seq::*` is doing, it is useful to look at the output of, say:

```
match /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*
```

which would look something like this:

```
/files/var/lib/pgsql/12/data/pg_hba.conf/1
/files/var/lib/pgsql/12/data/pg_hba.conf/2
/files/var/lib/pgsql/12/data/pg_hba.conf/3
etc
```

Note the _absence_ of the `#comment` nodes. `seq::*` only matches nodes which consist only of digits.

To use it in a `set` command you could use something like this:

```
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[address="192.0.2.99/32"]/type host
```

which would either change the `type` of an existing entry to `host` if it already has an address of `192.0.2.99/32` or create a new numbered node if no such entry exists.

`seq::*` is not idempotent of itself, and the above example is not idempotent.

But we can use it to create an idempotent sequence of entries, like this

```
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[address="192.0.2.99/32"]/type host
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[type="host" and (address="192.0.2.99/32" or count(address)=0)]/database all
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[type="host" and (address="192.0.2.99/32" or count(address)=0)]/user     hal9000
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[type="host" and (address="192.0.2.99/32" or count(address)=0)]/address  "192.0.2.99/32"
set /files/var/lib/pgsql/12/data/pg_hba.conf/seq::*[type="host" and (address="192.0.2.99/32" or count(address)=0)]/method   md5
```

Which will create a new entry allowing the user `hal9000` from the ip `192.0.2.99`  to access all databases with an md5 password.

If the _same sequence_ of `set` commands is applied again, no change is made. So this set of commands is idempotent.

Note that if there a more entries with the same IP address `192.0.2.99/32` then this example is unlikely to work as-is, and a more sophisticated filter-expression would be required.

The `pg_hba.conf` file often contains a lot of entries with similar values. This make it harder to write filter-path expressions which are idempotent.

src/pathx.c
```diff
@@ -2891,6 +2912,25 @@ int pathx_expand_tree(struct pathx *path, struct tree **tree) {
...
+/* Generate a numeric string to use for step->name
+ * Scan tree->children for the highest numbered label, and add 1 to that
+ */
+static char *step_seq_choose_name(struct tree *tree) {
+    int max_node_n=0;
+    char *step_name;
+    for(tree=tree->children; tree!=NULL; tree=tree->next) {
+        if ( tree->label == NULL)
+          continue;
+        int node_n=atoi(tree->label);
```
lutter, line 2924: AFAIK, `atoi` should not be used anymore, and `strtoul` would be more appropriate. If you do that, check for overflow

lutter (approved):
Nice! I left a comment on using `atoi`, but other than that, really nice addition. And I am surprised by how little code it took.

src/pathx.c
georgehansper: I've updated the code to use `strtoul()` instead of `atoi()` and report an overflow if it happens.

Can you please re-approve this Pull Request with the change?

The diff is:
```diff
diff --git a/src/pathx.c b/src/pathx.c
index 4d59d1d4..f2f9c09e 100644
--- a/src/pathx.c
+++ b/src/pathx.c
@@ -35,7 +35,7 @@ static const char *const errcodes[] = {
     "no error",
     "empty name",
     "illegal string literal",
-    "illegal number",
+    "illegal number",                             /* PATHX_ENUMBER */
     "string missing ending ' or \"",
     "expected '='",
     "allocation failed",
@@ -2832,7 +2832,7 @@ static int locpath_search(struct locpath_trace *lpt,
     return result;
 }
 
-static char *step_seq_choose_name(struct tree *tree);
+static char *step_seq_choose_name(struct pathx *path, struct tree *tree);
 
 /* Expand the tree ROOT so that it contains all components of PATH. PATH
  * must have been initialized against ROOT by a call to PATH_FIND_ONE.
@@ -2883,7 +2883,7 @@ int pathx_expand_tree(struct pathx *path, struct tree **tree) {
         if (s->axis != CHILD && s->axis != SEQ)
             goto error;
         if (s->axis==SEQ && s->name == NULL) {
-            s->name = step_seq_choose_name(parent);
+            s->name = step_seq_choose_name(path, parent);
         }
         if (s->name == NULL )
             goto error;
@@ -2914,18 +2914,28 @@ int pathx_expand_tree(struct pathx *path, struct tree **tree) {
 
 /* Generate a numeric string to use for step->name
  * Scan tree->children for the highest numbered label, and add 1 to that
+ * numeric labels may be interspersed with #comment or other labels
  */
-static char *step_seq_choose_name(struct tree *tree) {
-    int max_node_n=0;
+static char *step_seq_choose_name(struct pathx *path, struct tree *tree) {
+    unsigned long int max_node_n=0;
+    unsigned long int node_n;
     char *step_name;
+    char *label_end;
     for(tree=tree->children; tree!=NULL; tree=tree->next) {
         if ( tree->label == NULL)
           continue;
-        int node_n=atoi(tree->label);
+        node_n=strtoul(tree->label, &label_end, 10);
+        if ( label_end == tree->label || *label_end != '\0' )
+          /* label is not a number - ignore it */
+          continue;
+        if ( node_n >= ULONG_MAX ) {
+          STATE_ERROR(path->state, PATHX_ENUMBER);
+          return NULL;
+        }
         if( node_n > max_node_n )
             max_node_n = node_n;
     }
-    if (asprintf(&step_name,"%d",max_node_n+1) >= 0)
+    if (asprintf(&step_name,"%lu",max_node_n+1) >= 0)
         return step_name;
     else
         return NULL;
```

georgehansper:
This Pull Request cannot be merged because the travis-ci checks are not running.

This all affects all other Pull Requests too, and the entire Augeas project is stalled until this problem is resolved.

MatthewHannigan:
Probably way too late, but this reminded my of the use of of `iota` in golang and c++ to indicate the next sequential number.

Issue #540, opened 2018-01-19
augtool: dump-json support?


examon:
@lutter 

Hi, I want to ask if it would be possible to add support for json output.
It would help me and my team a lot if there was a dump-json command for augtool in addition to the dump-xml.

Right now, we are transforming XML output that augtool produces to the matching json.
The problem is that this transformation takes relatively long time (because of fnmatch, see https://github.com/leapp-to/leapp-actors/blob/master/src/actors/common/augeas/augeas.py#L53).

Thank you

lutter:
I'd be very interested in converting the Augeas tree to JSON. The tricky part is what exactly the JSON representation would look like - it would be easy enough to do a direct translation of the tree into JSON, i.e. each node in the tree would become an object
```
  { "label": <label>
    "value": <value>
    "children: [ <node>, <node>, ... ] }
```

I have been hoping to get to a more compact representation, something where each level in the tree becomes a JSON object; e.g., something like the tree at `/files/etc/hosts/1` might become
```
  { "ipaddr": "127.0.0.1",
    "canonical": "localhost",
    "alias": [ "localhost.localdomain", "local.example.com" ] }
```
but such a representation would lose ordering (since the keys in a JSON object are not ordered) and it's not clear how you deal, e.g., with `#comment` nodes interspersed with other nodes. It's also hard to produce output that is stable, i.e., where the form of the output is the same for a given file, regardless of what a file contains. As an example, you might turn the options from `/etc/fstab` into an array of strings and that will work until one of those options has a value. At that point, each option would need to become an object.

Looking at your code, I see some other ways in which it could be sped up:

- switch to using the Python bindings to libaugeas rather than invoking it from the command line; right now, both `get_lenses` and `process_augeas_data` run augtool, and have it load everything which can be very slow
- when you use the bindings, you can call `aug_to_xml` directly, which returns an `xmlNode`, avoiding parsing the XML text
- you can find what lenses were used for a given file `file` by doing a `get /augeas/files/<file>/lens`
- similarly, to find all the files that were loaded with the lens `Mod.lns`, you'd do `match /augeas/files//path[../lens = "Mod.lns" or ../lens = "@Mod"]`
- if you want to look at a really fast way to iterate over the tree, have a look at how [`augmatch`](https://github.com/hercules-team/augeas/blob/master/src/augmatch.c) does it - it should be possible to emulate that technique in Python, though that might be overkill, and would require the not-yet-released Augeas 1.10 ;)

[...]

PR #717, opened 2021-03-05
Add JSON export to Augeas API and CLI

adamcritchley:
This pull request adds JSON output to Augeas and addresses issue https://github.com/hercules-team/augeas/issues/540.

Added a command dump-json to the CLI.
Added aug_to_json to the API with respective tests in test-api.c
Added json-c dependency to Debian/Redhat packages.
Added json-c to Docker build.

Here's a sample output when used on the CLI...
```bash
augtool> dump-json /files/etc/sudoers
{
  "augeas":{
    "match":"\/files\/etc\/sudoers",
    "sudoers":{
      "path":"\/files\/etc\/sudoers",
      "#comment":{
        "value":"See sudoers(5) for more information on \"#include\" directives:"
      },
      "Defaults":{
        "secure_path":{
          "value":"\"\/usr\/local\/sbin:\/usr\/local\/bin:\/usr\/sbin:\/usr\/bin:\/sbin:\/bin:\/snap\/bin\""
        }
      },
      "spec":{
        "user":{
          "value":"%sudo"
        },
        "host_group":{
          "host":{
            "value":"ALL"
          },
          "command":{
            "value":"ALL",
            "runas_user":{
              "value":"ALL"
            },
            "runas_group":{
              "value":"ALL"
            }
          }
        }
      },
      "#includedir":{
        "value":"\/etc\/sudoers.d"
      }
    }
  }
}
```


adamcritchley:
Will be updating the pull request with a fix for the build.

raphink:
Interesting feature. Is there a reason for using JSONL instead of JSON? Unless I'm mistaken, JSONL is mainly useful to output streams of JSON objects without dealing with the final comma problem. But the objects you're exporting here are finished.

adamcritchley:
My apologies! Our team started with JSONL but eventually just settled with JSON to make it queryable through presto SQL statements.
https://prestodb.io/docs/current/functions/json.html

I'll change the names from JSONL to JSON. If you have any other recommendations then I'll make those changes today as well.

Does everything else look in order for this pull request?

raphink:
I'll let @lutter and @georgehansper comment on the implementation 😁

adamcritchley:
Hello @lutter and @georgehansper! Is there anything I can do to improve this PR? Thanks!

georgehansper:
Sorry, I haven't looked it it as yet. I will have a look shortly.

adamcritchley:
For context, here's an example SQL query that we can execute on the JSON data returned from Augeas.
```SQL
with dataset as (
  select CAST(json_extract(augeas_json, '$.augeas.sshd_config.PermitRootLogin.value') as VARCHAR) as perm from augeas_table
  ) select perm from dataset where perm is not null;```

**Comment by georgehansper** [comment]
I haven't looked at the code in detail as yet.

I have ran a couple of tests on this code, and I have a few comments.

**`\/`**
I notice though that the slashes of `match` and  `path` are escaped as `\/`
In fact, all strings have the `/` replaced by `\/'.
I think this is undesirable, as it would imply a simple string, rather than a path.

**`augeas`**
The whole JSON tree is wrapped up as a single object called `augeas`. I can't quite see the point of this. It just becomes something the receiving program needs to unwrap and discard.

**`match` and `path`**
The top-level JSON object has a few extra pieces of metadata added, `match` and `path`

These may or may not be required, depending on whether `dump-json` is trying dump 'matching nodes' or an entire subtree.

If `dump-json` is dumping 'matching nodes', then these are useful pieces of information.

If `dump-json` is dumping the entire subtree, then it would be better to represent the entire tree in JSON, eg

```
augtool> dump-json /files/etc/sudoers/Defaults
{
  "augeas":{
    "match":"\/files\/etc\/sudoers\/Defaults",
    "Defaults":{
      "path":"\/files\/etc\/sudoers\/Defaults[8]",
      "secure_path":{
        "value":"\/usr\/local\/sbin:\/usr\/local\/bin:\/usr\/sbin:\/usr\/bin:\/sbin:\/bin"
      }
    }
  }
}
```

would become:

```
augtool> dump-json /files/etc/sudoers/Defaults[last()]
{
  "files":{
    "etc":{
      "sudoers":{
        "Defaults":{
          "secure_path":{
            "value":"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
          }
        }
      }
    }
  }
}
```

The way `dump-json` works right now is analogous to the `print` command.
The `print` command prints the entire subtree of all matching nodes, and there is no reference to the original matching path statement.
eg `print /files/etc/hosts//alias` will print all the aliases in /etc/hosts, but doesn't tell you it's ignoring all the `ipaddr` and `canonical` nodes

I could argue that the person/program that issues the `dump-json` command should already know the original path-matching statement, and does not need `match` to know what it was.

Personally, I think a `dump-something` command should represent the original data-structure as closely as possible, so my own preference would be for the 2nd representation (above), and omit the metadata. Others may have different opinions on this.

This brings me to the bigger problem, and that is how to represent the Augeas tree in general.

Internally, the Augeas tree is a 'tree of lists', where each individual list-item has 3 (main) properties:
- label (mandatory)
- value (optional)
- list of child nodes (optional)

One important point here is that there is no requirement for either the label or value to be unique.

This shows itself in the `dump-json` output as "missing" nodes.

- /etc/sudoers can have multiple Defaults entries, and multiple #comments, but only the last one is printed (see your example).

- an entry in /etc/hosts can have multiple aliases, but only the last one is dumped, like this:

```
augtool> print /files/etc/hosts//alias
/files/etc/hosts/1/alias[1] = "localhost4"
/files/etc/hosts/1/alias[2] = "localhost.localdomain"
/files/etc/hosts/2/alias = "localhost6"
/files/etc/hosts/3/alias[1] = "www.example.com"
/files/etc/hosts/3/alias[2] = "ftp.example.com"
augtool> dump-json /files/etc/hosts//alias
{
  "augeas":{
    "match":"\/files\/etc\/hosts\/\/alias",
    "alias":{
      "path":"\/files\/etc\/hosts\/3\/alias[2]",
      "value":"ftp.example.com"
    }
  }
}
```


Lets consider a fictitious file: `/etc/auto.nfs` with the following contents:

```
# my comment
myserver     -ro,hard,intr       myserver.example.com:/mnt/bigdir
```

```
augtool> print /files/etc/auto.nfs
/files/etc/auto.nfs
/files/etc/auto.nfs/#comment = "my comment"
/files/etc/auto.nfs/1 = "myserver"
/files/etc/auto.nfs/1/opt[1] = "ro"
/files/etc/auto.nfs/1/opt[2] = "hard"
/files/etc/auto.nfs/1/opt[3] = "intr"
/files/etc/auto.nfs/1/location
/files/etc/auto.nfs/1/location/1
/files/etc/auto.nfs/1/location/1/host = "myserver.example.com"
/files/etc/auto.nfs/1/location/1/path = "/mnt/bigdir"
```

So the "natural" JSON representation of this might be:

```json
[
  {
    "label": "files",
    "children": [
      {
        "label": "etc",
        "children": [
          {
            "label": "auto.nfs",
            "children": [
              {
                "label": "#comment",
                "value": "my comment"
              },
              {
                "label": 1,
                "value": "myserver",
                "children": [
                  {
                    "label": "opt",
                    "value": "ro"
                  },
                  {
                    "label": "opt",
                    "value": "hard"
                  },
                  {
                    "label": "opt",
                    "value": "intr"
                  },
                  {
                    "label": "location",
                    "children": [
                      {
                        "label": 1,
                        "children": [
                          {
                            "label": "host",
                            "value": "myserver.example.com"
                          },
                          {
                            "label": "path",
                            "value": "/mnt/bigdir"
                          }
                        ]
                      }
                    ]
                  }
                ]
              }
            ]
          }
        ]
      }
    ]
  }
]
```

This style would work with any existing config-file and lens.

However, it is neither compact nor readable. The current implementation of `dump-json` makes for a nice, readable format.

There are a couple of possible solutions.

The most obvious is to do what Augeas already does, which is to append an index to duplicated labels.

If there is only one comment in the file, it appears as:

```
/files/etc/auto.nfs/#comment "my comment"
```

If there are multiple, it appears as

```
/files/etc/auto.nfs/#comment[1] "my comment"
/files/etc/auto.nfs/#comment[2] "my other comment"
```

------

This PR implements dump-json as a augtool CLI command, but does not offer this as an API.
That may be out-of-scope for this PR, but it is something to think about.

Since `dump-json` is currently "only" an interactive command in `augtool`, it would be possible to take it out of `augtool` altogether, and make it a stand-alone program. I'm not sure what others think about this. There are pros and cons.

adamcritchley:
@georgehansper, sorry for the delay we're currently reviewing your request. I'll be in touch with our decision shortly. Thanks!

adamcritchley:
Hello @raphink and @georgehansper! After internal review, we decided not to pursue this PR further as our use case will require enough modifications that it will differ from Augeas' original design. However, we will continue to keep our fork open source. Thanks!

PR #720, opened 2021-04-06
mongodb (bug): Fix paths for MongoDB assets

brianredbeard:
A number of files used by MongoDB had path changes, moving from the name
`mongodb.<extension>` to `mongod.<extension>` (changing "db" -> "d").

This updates those files (and their corresponding tests) to reflect the
correct paths, as of MongoDB v4.4.

Special thanks to Frank Büttner of Max-Delbrück-Centrum für Molekulare
Medizin for bringing this to my attention and noting the change.
As the idiom goes: "Many hands make light work"

### DIFF
--- a/lenses/mongodbserver.aug
+++ b/lenses/mongodbserver.aug
@@ -1,6 +1,6 @@
 (*
 Module: MongoDBServer
-  Parses /etc/mongodb.conf
+  Parses /etc/mongod.conf
 
 Author: Brian Redbeard <redbeard@dead-city.org>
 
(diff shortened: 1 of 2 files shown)


brianredbeard:
Given the date it was written (May 8, 2013) I would say it was likely written for Mongo 2.x.  To be fair, when I got the email from the user asking for the update ~18 months ago I was surprised that anyone was still using it.  The lenses were created in order to install OpenShift in an era before Kubernetes even existed.

If no one is using them and they're broken, it's probably even better to remove the lenses from that PR.

Issue #719, opened 2021-04-05
NetworkManager new values are added using double quotes which looks wrong according to NM


teclator:
Apparently since this change https://github.com/hercules-team/augeas/commit/bf25803ba4b3aa4e81feefb8fc217753ab6005c4 new entries are added using double quotes which NetworkManager does not seem to support.

Editing an entry works fine but it is not the case of new entries:

```
$ sudo augtool set /files/etc/NetworkManager/system-connections/enp0s25.nmconnection/ipv4/dns "8.8.8.8;8.8.4.4"
Saved 1 file(s)

$ cat /etc/NetworkManager/system-connections/enp0s25.nmconnection | grep ^dns=
dns="8.8.8.8;8.8.4.4"

$ cat /etc/NetworkManager/system-connections/enp0s25.nmconnection | grep ^method
method=auto
method=auto

$ sudo augtool rm /files/etc/NetworkManager/system-connections/enp0s25.nmconnection/ipv4/method
rm : /files/etc/NetworkManager/system-connections/enp0s25.nmconnection/ipv4/method 1
Saved 1 file(s)

$ sudo augtool set /files/etc/NetworkManager/system-connections/enp0s25.nmconnection/ipv4/method auto
Saved 1 file(s)

$ sudo augtool set /files/etc/NetworkManager/system-connections/enp0s25.nmconnection/ipv6/method manual
Saved 1 file(s)

$ cat /etc/NetworkManager/system-connections/enp0s25.nmconnection | grep ^method
method="auto" 
method=manual

```

And from the journal logs we can see:

```

<warn>  [1617622626.9161] keyfile: ipv4.dns: ignoring invalid DNS server IPv4 address '"8.8.8.8'
<warn>  [1617622626.9161] keyfile: ipv4.dns: ignoring invalid DNS server IPv4 address '8.8.4.4"'
<warn>  [1617622626.9162] settings: load: failure to load "/etc/NetworkManager/system-connections/enp0s25.nmconnection": invalid connection: ipv4.method: property is invalid

```

teclator:
@raphink could you take a look?, we started to use the Desktop lense but it has problems when '#' is used in the values that is basically what your [commit](https://github.com/hercules-team/augeas/commit/bf25803ba4b3aa4e81feefb8fc217753ab6005c4) fixed.

teclator:
@mchf thnx for taking care!

PR #723, opened 2021-05-26
Modified NetworkManager lense

mchf:
The lense will not encapsulate value of a newly created entry by quotes.
Values containing # are accepted (allowed character e.g. for WPA-PSK)

## Details ##

* For WPA-PSK each character in the pass-phrase must have an encoding in the range of 32 to 126 (decimal), inclusive. (IEEE Std. 802.11i-2004, Annex H.4.1)

E.g. # and space character are included in this range.

* adresses https://github.com/hercules-team/augeas/issues/719

### DIFF
--- a/lenses/networkmanager.aug
+++ b/lenses/networkmanager.aug
@@ -34,13 +34,17 @@ let eol        = Util.eol
  *                        ENTRY
  * GLib entries can contain semicolons, entry names can contain spaces and
  * brackets
+ *
+ * At least entry for WPA-PSK definition can contain all printable ASCII
+ * characters including '#', ' ' and others. Comments following the entry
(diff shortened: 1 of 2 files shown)


raphink:
Thanks for this. Could you add tests to identify the current issue and validate the fix please?

mchf:
I've added some test for the ```get``` direction. However, I haven't been able to create a test for writting new value (I wanted to check that new value is written without quotes) but I haven't found out how it could be done. But keep trying.

mchf:
ok, even ```put``` direction is now tested. It covers the issue https://github.com/hercules-team/augeas/issues/719 referenced in the description

raphink (approved):
LGTM

PR #724, opened 2021-06-22
Support negative integers in sudoers lens

anroots-tw:
[timestamp_timeout](https://help.ubuntu.com/community/RootSudoTimeout) in `/etc/sudoers` can have a negative value, but current augeas lens definition doesn't support this.

Modify lens regex to also match negative integers.

Issue is similar to https://libvir-list.redhat.narkive.com/gFPonY7Z/libvirt-patch-adapt-augeas-profile-to-handle-negative-int-values

### DIFF
--- a/lenses/sudoers.aug
+++ b/lenses/sudoers.aug
@@ -153,7 +153,7 @@ let sto_to_spc = store /[^", \t\n\\]+|"[^", \t\n\\]+"/
 let sto_to_spc_no_dquote = store /[^",# \t\n\\]+/ (* " relax emacs *)
 
 (* Variable: sto_integer *)
-let sto_integer = store /[0-9]+/
+let sto_integer = store /-?[0-9]+/
 
 
 (* Group: Comments and empty lines *)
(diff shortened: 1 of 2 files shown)


raphink:
Thanks. Could you add a unit test for this case please?

anroots-tw:
I spent a few hours trying to learn the project more deeply and set up for lens test development, without success (I'm in over my head a bit).

_(Note to self: put more time here when able)_

raphink:
From the project root, you simply need to launch  

```
augparse -I lenses lenses/tests/test_sudoers.aug
```

Once you have this setup, you can add a test in `test_sudoers.aug` to check for negative integers and verify that it still passes.

anroots-tw:
Thanks, that helped. I was stuck on trying to compile/run augeas from git; but that wasn't needed - APT version of `augeas-tools` provided working augparse for tests.

Proposed two test cases in  bf1c6d0

raphink:
Hmmm, Travis CI has stopped building. I've checked the patch on my machine and I'll merge.

We really need to migrate to another CI.

PR #737, opened 2021-10-12
Static analysis code review

cumamon-san:
Hi, I have a static analysis report for augeas 1.12.0 and propose these changes. Please, let me know if I should do something differently.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -1452,7 +1452,7 @@ int aug_match(const struct augeas *aug, const char *pathin, char ***matches) {
         if (*matches != NULL) {
             for (i=0; i < cnt; i++)
                 free((*matches)[i]);
-            free(*matches);
+            FREE(*matches);
         }
     }
     free_pathx(p);
--- a/src/fa.c
+++ b/src/fa.c
@@ -1851,10 +1851,12 @@ struct fa *fa_make_basic(unsigned int basic) {
         return fa_make_epsilon();
     } else if (basic == FA_TOTAL) {
         struct fa *fa = fa_make_epsilon();
-        r = add_new_trans(fa->initial, fa->initial, UCHAR_MIN, UCHAR_MAX);
-        if (r < 0) {
-            fa_free(fa);
-            fa = NULL;
+        if (fa) {
+            r = add_new_trans(fa->initial, fa->initial, UCHAR_MIN, UCHAR_MAX);
+            if (r < 0) {
+                fa_free(fa);
+                fa = NULL;
+            }
         }
         return fa;
     }
--- a/src/lens.c
+++ b/src/lens.c
@@ -651,6 +651,7 @@ struct value *lns_make_prim(enum lens_tag tag, struct info *info,
 
     return make_lens_value(lens);
  error:
+    unref(lens, lens);
     return exn;
 }
 
--- a/src/lens.c
+++ b/src/lens.c
@@ -677,6 +677,8 @@ static struct value *disjoint_check(struct info *info, bool is_get,
         goto done;
 
     fa = fa_intersect(fa1, fa2);
+    if (! fa)
+        goto done;
     if (! fa_is_basic(fa, FA_EMPTY)) {
         size_t xmpl_len;
         char *xmpl;
@@ -1646,7 +1648,8 @@ static void rtn_rules(struct rtn *rtn, struct lens *l) {
         /* cfg(l1 . l2 ... ln, N) -> N := N1 . N2 ... Nn */
         for (int i=0; i < l->nchildren-1; i++) {
             struct state *s = add_state(prod);
-            RTN_BAIL(rtn);
+            if (s == NULL)
+                goto error;
             add_trans(rtn, start, s, l->children[i]);
             RTN_BAIL(rtn);
             start = s;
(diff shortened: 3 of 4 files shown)


src/lens.c
```diff
@@ -1646,7 +1649,8 @@ static void rtn_rules(struct rtn *rtn, struct lens *l) {
         /* cfg(l1 . l2 ... ln, N) -> N := N1 . N2 ... Nn */
         for (int i=0; i < l->nchildren-1; i++) {
             struct state *s = add_state(prod);
-            RTN_BAIL(rtn);
+            if (s == NULL)
+                goto error;
```
g0hl1n, line 1653: IMHO this changes behavior. Can you please describe why it makes sense to replace RTN_BAIL(rtn) with checking s for NULL?

src/lens.c
```diff
@@ -1664,7 +1668,8 @@ static void rtn_rules(struct rtn *rtn, struct lens *l) {
     case L_STAR: {
         /* cfg(l*, N) -> N := N . N' | eps */
         struct state *s = add_state(prod);
-        RTN_BAIL(rtn);
+        if (s == NULL)
+            goto error;
```
g0hl1n, line 1672: IMHO this changes behavior. Can you please describe why it makes sense to replace RTN_BAIL(rtn) with checking s for NULL?

g0hl1n (changes requested):
Basically looks good to me. I just would like to see some explanation on the RTN_BAIL changes in lens.c

src/lens.c
KucherenkoAO: Because further in line `add_trans (rtn, s, prod-> end, l-> child);` `s` is being dereferenced inside `add_trans`. I'm not sure if `RTN_BAIL (rtn)` protects against this.

raphink:
LGTM, but I'd like @lutter and @georgehansper 's opinions

src/lens.c
georgehansper: Near as I can tell, in the event that memory allocation fails, this is detected within `add_state()` and the error code is stored in `prod->lens->info->error`
According to the comment in src/info.h:

/* There is only one struct error for each Augeas instance */

RTN_BAIL(rtn) checks rtn->info->error and executes `goto error;` on any error.
So the case of `s == NULL` is covered by the pre-existing `RTN_BAIL(rtn)` code.

That's probably beyond the capabilities of the static code analyser to evaluate.
I think the existing code here should stay unchanged.

src/syntax.c
```diff
@@ -1505,7 +1507,7 @@ static struct value *compile_concat(struct term *exp, struct ctx *ctx) {
 
     struct type *t = exp->type;
     struct info *info = exp->info;
-    struct value *v;
+    struct value *v = NULL;
```
georgehansper, line 1510: v is assigned in every branch of the subsequent if-else-if... structure, so this change in conjunction with removing line 1549 (below) is unneccessary

I would leave this unchanged. because it's more a question of style than making an improvement.

src/lens.c
georgehansper: Same as for 1649 above

src/augeas.c
```diff
@@ -1452,7 +1452,7 @@ int aug_match(const struct augeas *aug, const char *pathin, char ***matches) {
         if (*matches != NULL) {
             for (i=0; i < cnt; i++)
                 free((*matches)[i]);
-            free(*matches);
+            FREE(*matches);
```
georgehansper, line 1455: OK

src/lens.c
```diff
@@ -651,6 +651,7 @@ struct value *lns_make_prim(enum lens_tag tag, struct info *info,
 
     return make_lens_value(lens);
  error:
+    unref(lens, lens);
```
georgehansper, line 654: OK

src/fa.c
```diff
@@ -1851,10 +1851,12 @@ struct fa *fa_make_basic(unsigned int basic) {
...
-        r = add_new_trans(fa->initial, fa->initial, UCHAR_MIN, UCHAR_MAX);
-        if (r < 0) {
-            fa_free(fa);
-            fa = NULL;
+        if (fa) {
+            r = add_new_trans(fa->initial, fa->initial, UCHAR_MIN, UCHAR_MAX);
+            if (r < 0) {
+                fa_free(fa);
+                fa = NULL;
+            }
```
georgehansper, line 1859: OK

src/syntax.c
```diff
@@ -414,8 +414,10 @@ struct value *make_exn_value(struct info *info,
         return NULL;
 
     v = make_value(V_EXN, ref(info));
-    if (ALLOC(v->exn) < 0)
+    if (ALLOC(v->exn) < 0) {
+        unref(v, value);
         return info->error->exn;
+    }
```
georgehansper, line 420: Seems OK to me

src/syntax.c
```diff
@@ -2076,6 +2078,7 @@ int interpreter_init(struct augeas *aug) {
             p += 1;
         q = strchr(p, '.');
         name = strndup(p, q - p);
+        ERR_NOMEM(name == NULL, aug);
```
georgehansper, line 2081: OK

src/syntax.c
```diff
@@ -1514,8 +1516,11 @@ static struct value *compile_concat(struct term *exp, struct ctx *ctx) {
         const char *s2 = v2->string->str;
         v = make_value(V_STRING, ref(info));
         make_ref(v->string);
-        if (ALLOC_N(v->string->str, strlen(s1) + strlen(s2) + 1) < 0)
-            goto error;
+        if (ALLOC_N(v->string->str, strlen(s1) + strlen(s2) + 1) < 0) {
+            unref(v->string, string);
+            unref(v, value);
+            return exp->info->error->exn;
+        }
```
georgehansper, line 1523: Seems OK to me

PR #738, opened 2021-10-13
Allow Redis lens to set `SAVE ""` as a valid option

mhagstrand:
Before Redis 6.2 removing all the `SAVE` directives in the config file would disable snapshots. 
https://github.com/redis/redis/blob/6.0/redis.conf#L299-L305

In Redis 6.2 removing all the `SAVE` directives in the config file will result in having snapshots enabled.
https://github.com/redis/redis/blob/6.2/redis.conf#L369-L377

To disable snapshots in Redis 6.2 requires setting `SAVE ""` as a line in the config file. This change allows the Redis lens to use `SAVE ""` a valid directive in the Redis configuration. Without this change, snapshots cannot be disabled using this lens.

Using `SAVE ""` to disable snapshots has been valid syntax since Redis 2.8
https://github.com/redis/redis/blob/2.8/redis.conf#L141-L145

### DIFF
--- a/lenses/redis.aug
+++ b/lenses/redis.aug
@@ -65,12 +65,17 @@ let standard_entry =
 let save = /save/
 let seconds = [ label "seconds" . Quote.do_quote_opt_nil (store Rx.integer) ]
 let keys = [ label "keys" . Quote.do_quote_opt_nil (store Rx.integer) ]
+let save_val =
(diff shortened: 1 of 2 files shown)


mhagstrand:
@raphink is there anything else this needs for it be to merged?

PR #745, opened 2022-01-11
Chrony: add new directives and options

mlichvar:
This adds support for new configuration added in chrony 4.1 and 4.2.

### DIFF
--- a/lenses/chrony.aug
+++ b/lenses/chrony.aug
@@ -57,8 +57,11 @@ module Chrony =
 (************************************************************************
  * Group: Create required expressions
  ************************************************************************)
+    (* Variable: hex *)
+    let hex = /[0-9a-fA-F]+/
+
     (* Variable: number *)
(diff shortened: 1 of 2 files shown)


lenses/chrony.aug
```diff
@@ -57,8 +57,11 @@ module Chrony =
 (************************************************************************
  * Group: Create required expressions
  ************************************************************************)
+    (* Variable: hex *)
+    let hex = /[0-9a-fA-F]+/
+
     (* Variable: number *)
-    let number = integer | decimal | decimal . /[eE]/ . integer
+    let number = integer | decimal | decimal . /[eE]/ . integer | hex
```
raphink, line 64: You can use  `Rx.hex` (from `rx.aug`) and then you don't need to define it.

```suggestion
    let number = integer | decimal | decimal . /[eE]/ . integer | Rx.hex
```

raphink (approved):

PR #747, opened 2022-02-02
Replace deprecated 'security_context_t' with 'char *'

Leo-Schmit:
https://github.com/SELinuxProject/selinux/commit/7a124ca275
https://github.com/SELinuxProject/selinux/pull/212/files

### DIFF
--- a/src/transform.c
+++ b/src/transform.c
@@ -918,7 +918,7 @@ static int transfer_file_attrs(FILE *from, FILE *to,
     struct stat st;
     int ret = 0;
     int selinux_enabled = (is_selinux_enabled() > 0);
-    security_context_t con = NULL;
+    char *con = NULL;
 
     int from_fd;
     int to_fd = fileno(to);


georgehansper:
This is in line with the comments in `/usr/include/selinux/selinux.h` (libselinux-devel-3.3-4.fc36)

```C
/* No longer used; here for compatibility with legacy callers. */
typedef char *security_context_t
#ifdef __GNUC__
__attribute__ ((deprecated))
#endif
;
```

`transform.c` uses the variable `con`as an argument to the libselinux functions `fgetfilecon()` and  `fsetfilecon()`
These were updated in 2014 to use `char **` instead of `security_context_t *`
https://github.com/SELinuxProject/selinux/commit/9eb9c9327563014ad6a807814e7975424642d5b9

georgehansper (approved):
As per https://github.com/hercules-team/augeas/pull/747#issuecomment-1242697222

PR #752, opened 2022-03-14
Add augprint tool for creating idempotent augtool scripts

georgehansper:
This progam will produce an augtool script capable of recreating a file, using path-expressions which are independent of the absolute position number [1] [2] [3]...  eg

```
set /files/etc/hosts/seq::*[ipaddr='127.0.0.1']/ipaddr '127.0.0.1'
set /files/etc/hosts/seq::*[ipaddr='127.0.0.1']/canonical 'localhost'
set /files/etc/hosts/seq::*[ipaddr='127.0.0.1']/alias[.='localhost.localdomain'] 'localhost.localdomain'
```

The position-indepentent path-expressions make it possible to copy-paste selected sections of the output, and use them "as-is" with augtool or other Augeas clients.

This tool is implemented as a 'client' of the existing Augeas API, and this Pull Request makes no changes to the augeas library or API.

I was going call this tool `augprint` , but there are a number of different ways this kind of idempotency can be achieved, so augprint seemed a little strong - these path-expressions should be considered a "suggestion".

I am happy to entertain other names for this tool, eg  augtool-gen  augtool-make


georgehansper:
I've updated this Pull Request to change the name of the program from augsuggest -> augprint
Mostly because it is easier to remember, and perhaps a better reflection of the nature of the output.

georgehansper:
Unless there are any comments or objections, I will merge this on the 20th Nov 2022

PR #757, opened 2022-04-11
Systemd lens fixes

michalvasko:
Fixes of `systemd` lens as described in the commit messages, tests included.


igalic (approved):
this seems sensible

georgehansper (approved):
systemd accepts empty environment variables
systemd accepts and ignores leading spaces before the command (ExecStartPre ExecStart ExecStartPost)

This change works as described, all tests passing.

PR #758, opened 2022-04-20
semanage: Fix parsing of ignoredirs

rwmjones:
From /etc/selinux/semanage.conf from a RHEL 9.1 system, this line
caused problems:

  ignoredirs=/root;/bin;/boot;/dev;/etc [...]

Parse this as a list of Rx.fspath.

Also this adds the RHEL 9 file as another test case and adjusts the
output of the existing test case.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2077120

### DIFF
--- a/lenses/semanage.aug
+++ b/lenses/semanage.aug
@@ -23,7 +23,12 @@ let sep = IniFile.sep "=" "="
 let empty = IniFile.empty
 let eol = IniFile.eol
 
-let entry = IniFile.entry IniFile.entry_re sep comment
+let list_keys = "ignoredirs"
+let scl = del ";" ";"
+let fspath = /[^ \t\n;#]+/ (* Rx.fspath without ; or # *)
+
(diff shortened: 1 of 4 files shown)


georgehansper (commented):
This fix works, and the solution is in-line with the existing Inifile module.

In the event of someone using the previous version of the `semange.aug` lens to modify an `ignoredirs` entry with just a single directory, that person would need to update their Augeas path from (eg. for a value of `/root` ):
```
set /files/etc/selinux/semanage.conf/ignoredirs "/root"
```
to
```
set /files/etc/selinux/semanage.conf/ignoredirs/1 "/root"
```

Given that the previous version of the `semange.aug` lens did not allow `;` in the `ignoredirs` value, this use-case is likely to be rare.

rwmjones:
(rebased on top of master, no other changes)

georgehansper (approved):
As per my previous comment, I think this is good solution.

PR #759, opened 2022-04-21
Key may be quoted

towo:
The lens assumed an unquoted word-like key, while it may be unquoted.

Technically, according to the
[spec](https://github.com/toml-lang/toml/blob/cbf3b13128fb717517afbd22f8fcb665a0b0b035/toml.abnf#L73),
it even allows for a few more non-word chars.

Test not adjusted (yet) because the format confused me.

### DIFF
--- a/lenses/toml.aug
+++ b/lenses/toml.aug
@@ -119,7 +119,10 @@ let inline_table (value:lens) = [ label "inline_table" . lbrace
                    . ( (Build.opt_list (entry_base value) comma . space_or_empty? . rbrace)
                       | rbrace ) ]
 
-let entry = [ label "entry" . Util.indent . store Rx.word . Sep.space_equal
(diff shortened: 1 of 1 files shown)


towo:
Weirdly enough, `augmatch` is also giving the output keys *unquoted* whilst `augtool` is giving the output keys *quoted*, which affects matching.

georgehansper:
This change fails testing with the message:
```
/home/george/github/augeas/lenses/toml.aug:125.0-126.88:Failed to compile entry
/home/george/github/augeas/lenses/toml.aug:125.44-.70:exception: overlapping lenses in tree union.put
    Example matched by both: 
    First lens: /home/george/github/augeas/lenses/toml.aug:122.17-.60:
    Second lens: /home/george/github/augeas/lenses/toml.aug:123.19-.32:
```

I would also suggest that, if the Toml lens is to support quoted keys, too, then this should be included in a test-case.

PR #767, opened 2022-07-25
[FIX] Copied all entries from sudoers doco and added functionality

nebakke:
On a system implementing 'Defaults iolog_dir' - augeas fails to parse sudoers as the options goes unrecognised.
By extension, implemting LOG_OUTPUT and LOG_INPUT tags causes the same sort of parsing issue.

Extended with the required options and any other updates, as per the sudoers man page.


georgehansper:
This change currently fails testing.

The problem appears to be around the keyword 'type'  in `let parameter_string_nobool_kw = ` ...

Issue #778, opened 2022-11-03
Bug Report:  null-pointer dereference in function fa_expand_nocase() (and so on) at src/fa.c 


ckxckx:
We find a crash report with [harness](https://github.com/google/oss-fuzz/blob/master/projects/augeas/augeas_fa_fuzzer.cc). 

Details can be seen from image below ([vul location](https://github.com/hercules-team/augeas/blob/e9c7ce30a54787ca85916322d3d06eeff6e74d1a/src/fa.c#L4449)) :
![image](https://user-images.githubusercontent.com/18208119/199663745-49c1de25-75c8-47ea-b6f7-a2aa3a02cc5f.png)

Download [PoC](https://raw.githubusercontent.com/ckxckx/sharesomething/b3b397b8c635959083b7ff4b84904d4055fab16d/augeas/crash-dd2a8067b5156541d297248bcfd84b61e95955a7) here.

Here is the snapshot in our testing:
![image](https://user-images.githubusercontent.com/18208119/199666925-a64be858-f84a-4590-83aa-76706ee0fb88.png)

Similar issues may also happen [here](https://github.com/hercules-team/augeas/blob/e9c7ce30a54787ca85916322d3d06eeff6e74d1a/src/fa.c#L4339) and [here](https://github.com/hercules-team/augeas/blob/e9c7ce30a54787ca85916322d3d06eeff6e74d1a/src/fa.c#L4371))

Looking forward to your validation and reply !!!

Issue #671, opened 2020-04-28
NULL pointer dereference in re_case_expand()


ltx2018:
we found NULL pointer dereference in re_case_expand() by fuzz testing.
```
==8==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000004 (pc 0x00000057f8d9 bp 0x7ffff56ac5f0 sp 0x7ffff56ac470 T0)
==8==The signal is caused by a READ memory access.
==8==Hint: address points to the zero page.
SCARINESS: 10 (null-deref)
    #0 0x57f8d8 in re_case_expand /src/augeas/src/fa.c
    #1 0x57f0b3 in fa_expand_nocase /src/augeas/src/fa.c:4453:9
    #2 0x553bc0 in LLVMFuzzerTestOneInput /src/augeas/augeas_fa_fuzzer.cc:52:10
    #3 0x459d81 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerLoop.cpp:556:15
    #4 0x4449a1 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:292:6
    #5 0x44a65e in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:774:9
    #6 0x4747d2 in main /src/llvm/projects/compiler-rt/lib/fuzzer/FuzzerMain.cpp:19:10
    #7 0x7f9df1dee82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #8 0x41dd58 in _start (/out/augeas_fa_fuzzer+0x41dd58)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /src/augeas/src/fa.c in re_case_expand
```

it can be reproduced by simply run test code:
```
int Size = 4;
char *new_str = (char *)malloc(Size + 1);
if (new_str == NULL)
     return 1;
new_str[0] = '{';
new_str[1] = '&';
new_str[2] = '.';
new_str[3] = '{';
new_str[Size] = '\0';
char *s;
size_t len;
fa_expand_nocase(new_str, Size, &s, &len);
```

[...]

PR #785, opened 2022-11-21
fix NULL pointer dereferences in fa_expand_nocase()

century6:
Fix the following issues:
https://github.com/hercules-team/augeas/issues/778
https://github.com/hercules-team/augeas/issues/671


georgehansper:
Is there some way to reproduce or demonstrate this problem?

Issue #760, opened 2022-04-27
Unable to compile in Termux


jose1711:
I am getting this trying to compile in Termux:
```
In file included from regex.c:68:
In file included from ./regex_internal.h:25:
In file included from ./stdio.h:43:
In file included from /data/data/com.termux/files/usr/include/stdio.h:47:
In file included from ./string.h:41:
In file included from /data/data/com.termux/files/usr/include/string.h:187:
/data/data/com.termux/files/usr/include/bits/fortify/string.h:91:47: error: use of undeclared identifier '__USE_FORTIFY_LEVEL'
        __clang_error_if(__bos_unevaluated_le(__bos(dst), __builtin_strlen(src)),
                                              ^
./cdefs.h:123:48: note: expanded from macro '__bos'
#define __bos(ptr) __builtin_object_size (ptr, __USE_FORTIFY_LEVEL > 1)
                                               ^
In file included from regex.c:68:
In file included from ./regex_internal.h:25:
In file included from ./stdio.h:43:
In file included from /data/data/com.termux/files/usr/include/stdio.h:47:
In file included from ./string.h:41:
In file included from /data/data/com.termux/files/usr/include/string.h:187:
/data/data/com.termux/files/usr/include/bits/fortify/string.h:91:47: error: use of undeclared identifier '__USE_FORTIFY_LEVEL'
```

```
$ gcc --version                                                                                                                  62% (CHG)─╯
clang version 14.0.1
Target: aarch64-unknown-linux-android24
Thread model: posix
InstalledDir: /data/data/com.termux/files/usr/bin
```

DimitryAndric:
It looks like Termux comes with its own C headers, as I see `/data/data/com.termux/files/usr/include/stdio.h`, `/data/data/com.termux/files/usr/include/string.h`, etc. And these seem to clash a bit with `./stdio.h`, `./string.h` and others, which appear to come from gnulib. Might be tricky to untangle.

PR #791, opened 2022-12-08
Only use GNU specific strerror_r() API when __GLIBC__ is defined.

DimitryAndric:
As reported in https://pkg-status.freebsd.org/gohan05/data/maini386PR265425-default/2022-12-06_12h24m32s/logs/errors/augeas-1.12.0_3.log, augeas uses a GNU specific `strerror_r()` invocation, even when compiling on FreeBSD. This is because gnulib's `regex.h` defines `__USE_GNU` when `_GNU_SOURCE` is used.

It is better to also check for `__GLIBC__`, since the GNU specific variant of `strerror_r()` is really a glibc invention. This will also make the code compile without warnings on Alpine Linux, which uses musl libc.

### DIFF
--- a/src/internal.c
+++ b/src/internal.c
@@ -431,8 +431,8 @@ char *cleanpath(char *path) {
 
 const char *xstrerror(int errnum, char *buf, size_t len) {
 #ifdef HAVE_STRERROR_R
-# ifdef __USE_GNU
-    /* Annoying linux specific API contract */
+# if defined(__USE_GNU) && defined(__GLIBC__)
+    /* Annoying GNU specific API contract */
     return strerror_r(errnum, buf, len);
 # else
     strerror_r(errnum, buf, len);


igalic:
this makes me wonder if https://github.com/hercules-team/augeas/issues/760 is also glibc only

smortex (approved):

igalic:
@georgehansper as per @smortex's review, this has been merged in FreeBSD's port system: https://github.com/freebsd/freebsd-ports/commit/48d51338852a1667227dd0735d03f77d72d67247

georgehansper:
Thanks for this PR and the comments. I will take a closer look in good time, to ensure there are no unwanted side-effects.

georgehansper:
I have downloaded FreeBSD 13.1 and compiled augeas 1.14.0 successfully from
* the tarball at https://github.com/hercules-team/augeas/releases/download/release-1.14.0/augeas-1.14.0.tar.gz
* a clone of this repository

Both of these compile without need of any modification to the source tree of augeas.

I have also downloaded the current FreeBSD 14.0 image (20221117), and successfully compiled augeas 1.14.0 from the above tarball, again, without any modification to the augeas source tree.

(FYI: The 1.12.0 tarball also compiles without modification on FreeBSD 13.1 and 14.0)

I cannot see any necessity for this PR.

Thoughts?

igalic:
FreeBSD CURRENT is in LLVM 14

The ports tree imports new versions and builds the tree with these versions.

DimitryAndric:
This was only in preparation for the import of llvm and clang 15.0 into FreeBSD. We do a few runs of the whole ports tree with the new compiler, linker and standard C++ library, to see if there any problems. 

One of the ports that resulted in a build error was augeas, specifically because it warned on the incorrect return value conversion (char* to int). Of course on the FreeBSD ports side, we could simply suppress this warning and continue as if everything was all right, but it seemed better to fix this properly.

That said, GNU libc is really the only C library that uses the `char *strerror_r(int, char *, size_t)` variant:
https://sourceware.org/git/?p=glibc.git;a=blob;f=string/string.h;h=54dd8344de25a2cdb3f4eac2ac4d0878d1b0d756;hb=HEAD#l422
Everybody else uses the POSIX variant, `int strerror_r(int, char *, size_t)`.

igalic:
thank you so much @DimitryAndric, that's a much more thorough explanation than i could have possibly mustered

thesamesam (commented):
Hit this downstream in Gentoo with musl as well.

PR #792, opened 2022-12-13
Remove `#include <malloc.h>`

ZhongRuoyu:
Hi! Submitting this patch to fix the build error in Homebrew/homebrew-core#118025.

`<malloc.h>` is not present on all platforms (e.g. macOS). To use `malloc`, `<stdlib.h>` should be the place to look for `malloc`, and it has already been `#include`d.

Signed-off-by: Ruoyu Zhong <zhongruoyu@outlook.com>

### DIFF
--- a/src/augprint.c
+++ b/src/augprint.c
@@ -69,6 +69,7 @@
 #include <string.h>
 #include <augeas.h>
 #include <errno.h>
+#include <libgen.h>        /* for basename() on FreeBSD and MacOS */
 #include <sys/param.h>     /* for MIN() MAX() */
 #include <unistd.h>
 #include "augprint.h"
--- a/src/augprint.c
+++ b/src/augprint.c
@@ -69,7 +69,6 @@
 #include <string.h>
 #include <augeas.h>
 #include <errno.h>
-#include <malloc.h>
 #include <sys/param.h>     /* for MIN() MAX() */
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


georgehansper (approved):
augprint compiles normally with this change, as per the description of this PR

georgehansper:
As it turns out, there is another similar problem affecting FreeBSD and brew, which causes `augprint -h` to fail
Can I ask that you update this PR to replace the malloc.h  line with:

```
#include <libgen.h>        /* for basename() on FreeBSD and MacOS */
```

igalic:
n.b.: including `libgen.h` on glibc systems brings [`basename(3)`](https://manpages.ubuntu.com/manpages/bionic/en/man3/basename.3.html)  to POSIX behaviour.

ZhongRuoyu:
> Can I ask that you update this PR to replace the malloc.h line...

OK, done. Tested locally and confirmed that this is fixed.

> n.b.: including `libgen.h` on glibc systems brings [`basename(3)`](https://manpages.ubuntu.com/manpages/bionic/en/man3/basename.3.html) to POSIX behaviour.

Thanks for bringing this up, @igalic. In this case I think it's fine since `argv[0]` is modifiable, but it's definitely worth some attention :)

georgehansper (approved):
Thanks for updating the PR

PR #798, opened 2023-01-09
lenses: Allow whitespace at the end of kernel commnd line

rwmjones:
Reported-by: Yongkui Guo
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2159282

### DIFF
--- a/lenses/cmdline.aug
+++ b/lenses/cmdline.aug
@@ -13,7 +13,7 @@ module Cmdline =
 
 let entry = [ key Rx.word . Util.del_str "=" . store Rx.no_spaces ] | [ key Rx.word ]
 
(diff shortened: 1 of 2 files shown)


igalic (commented):
do we not have tests for this?

rwmjones:
I'm not sure if you're asking if we test the feature in general (yes), or if we tested before whitespace at the end of a command line (no), or if I added a new test for this (also no) ?

igalic:
if we already have tests for the feature, then we should extend the tests, too

(i'm not going to ask people to add tests where there are none to begin with)

rwmjones:
The updated commit adds a test for a command line ending in whitespace.  Hope I did it right!

georgehansper (approved):
Works as described, includes a specific test

PR #801, opened 2023-01-28
Add Aptsources822 lens for Deb822 style Apt sources

### LINKED ISSUE #297: Add lens to handle Deb822 formatted Apt sources: /etc/apt/sources.d/*.sources
Apt supports a new format for describing its sources.  This format is based on RFC822 and is know as Deb822.  An example from the man page sources.list(5) is as follows:

```
Types: deb deb-src
URIs: http://example.com
Suites: stable testing
Sections: component1 component2
Description: short
 long long long
[option1]: [option1-value]

Types: deb
URIs: http://another.example.com
Suites: experimental
Sections: component1 component2
Enabled: no
Description: short
 long long long
[option1]: [option1-value]
```
Augeas should implement a new lens (perhaps sharing a common base with `debctrl` lens) that can parse this new format.  Users, at their choice, are allowed to pick any of the old or new formats.  This means that to be able to reliably modify sources list, programs must understand both the formats.

Motivation: Attackers may observe systems downloading a particular security update and realize that is vulnerable and attack it before the security update is applied.  Jacob Applebaum's talk at Debconf discusses this.  `apt-transport-tor` allows downloading packages anonymously via the Tor anonymity network.  This makes this kind of attacks much more difficult.  `apt-transport-tor` is available for Debian as a package.  To enable it, one has to update all the URLs in sources.list to look like `tor+http://` instead of `http://`.  This can't be done reliably if Augeas can parse only one of the two known file formats.

Further notes on Deb822: The current version of Apt supports both formats in the same files.  It interprets the file as one format or the other using a configuration setting.  This is a bit of a problem for Augeas as it parses files based on path/name.  The Apt developers have changed this behavior in Apt 1.1 which is available in Debian `experimental`.  In Apt 1.1:
 * `/etc/apt/sources.list` will always be old style format.
 * `/etc/apt/sources.list.d/*.list` will always be old format.
 * `/etc/apt/sources.list.d/*.sources` will always be new format files.

jvalleroy:
We are using this lens in FreedomBox to add/remove "tor+" prefix to Apt sources URIs. Original commit: https://salsa.debian.org/freedombox-team/freedombox/-/commit/cd21c7f247135828e63379cc6d32a8edcb53a582

Closes: #297.

Signed-off-by: James Valleroy <jvalleroy@mailbox.org>

### DIFF
--- a/lenses/aptsources.aug
+++ b/lenses/aptsources.aug
@@ -60,7 +60,7 @@ module Aptsources =
 
   (* View: filter *)
   let filter = (incl "/etc/apt/sources.list")
-      . (incl "/etc/apt/sources.list.d/*")
+      . (incl "/etc/apt/sources.list.d/*.list")
       . Util.stdexcl
 
   let xfm = transform lns filter
--- a/lenses/aptsources822.aug
+++ b/lenses/aptsources822.aug
@@ -1,0 +1,104 @@
+(*
+Module: Aptsources822
+  Augeas module for sources list (Deb822 style) for Apt package manager
+
+Authors:
+  James Valleroy <jvalleroy@mailbox.org>
+  Sunil Mohan Adapa <sunil@medhas.org>
+
+About: Reference
+  1. Deb822(5):
+     https://manpages.debian.org/bullseye/dpkg-dev/deb822.5.en.html
+  2. sources.list(5):
+     https://manpages.debian.org/bullseye/apt/sources.list.5.en.html
+
+About: License
+  This file is licensed under the LGPLv2+, like the rest of Augeas.
+
+About: Configuration files
+  This lens applies to files in /etc/apt/sources.list.d/ ending with the
+  extension .sources. See <filter>.
+
+*)
+
+
+module Aptsources822 =
+
+autoload xfm
+
+(* Variable: single_value_field_name
+   Names of known fields for which only a single value is allowed. *)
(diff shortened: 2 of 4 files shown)


georgehansper:
The lenses aptsources.aug and aptsources822.aug have overlapping definitions in the filter statement.
This results in an error because augeas can't determine which of the two lenses to apply. eg.

Error in /etc/apt/sources.list.d/man_examples.sources (mxfm_load)
  Lenses @Aptsources822 and @Aptsources could be used to load this file

According to the man page for sources.list.d:

> Filenames need to have either the extension .list or .sources depending on the contained format

This means that the existing lens should be limited to files named *.list, like this:

```
diff --git a/lenses/aptsources.aug b/lenses/aptsources.aug
index d7a6b3b1..e09634c7 100644
--- a/lenses/aptsources.aug
+++ b/lenses/aptsources.aug
@@ -60,7 +60,7 @@ module Aptsources =
 
   (* View: filter *)
   let filter = (incl "/etc/apt/sources.list")
-      . (incl "/etc/apt/sources.list.d/*")
+      . (incl "/etc/apt/sources.list.d/*.list")
       . Util.stdexcl

   let xfm = transform lns filter
```

Although this could be the subject of a separate PR, can I ask you to update this PR with the above change?

SunilMohanAdapa:
@georgehansper , Thank you for the review. I have collaborated with @jvalleroy in this MR and addressed all the review comments. Here are my changes:
* aptsources: Limit filtering one-line style sources to *.list
* Handle DOS style line terminators for empty, comment, stanza lines
* Allow leading, trailing and repeating empty lines
* Don't allow spaces and tabs in empty lines separating stanzas

Please let me know if further changes are required.

PR #805, opened 2023-02-17
Tmpfiles: few improvements

### LINKED ISSUE #795: tmpfiles lens: support for caret notation
Systemd introduced the caret notation, for credential extraction, in commit https://github.com/systemd/systemd/commit/e52f6f6358e5. In grandchild commit https://github.com/systemd/systemd/commit/a0f4426d0f12, the new feature was put to use in `tmpfiles.d/provision.conf`.

At commit 4f3bbeb3f782, augeas fails to parse the new type.

Please consider adding support for this caret notation. I assume the extension could look similar to commit fd681143cc49. Thanks!

Reference: https://bugzilla.redhat.com/show_bug.cgi?id=2155136

ptoscano:
- recognize `=`, `~`, `^` for the type specification (fixes #795)
- allow `:` as prefix for file modes

### DIFF
--- a/lenses/tmpfiles.aug
+++ b/lenses/tmpfiles.aug
@@ -50,11 +50,12 @@ Empty lines *)
 
   (* View: type
 One letter. Some of them can have a "+" and all can have an
-exclamation mark ("!") and/or minus sign ("-").
+exclamation mark ("!"), a minus sign ("-"), an equal sign ("="),
+a tilde character ("~") and/or a caret ("^").
 
 Not all letters are valid.
 *)
-  let type     = /([fFwdDevqQpLcbCxXrRzZtThHaAm]|[fFwpLcbaA]\+)!?-?/
+  let type     = /([fFwdDevqQpLcbCxXrRzZtThHaAm]|[fFwpLcbaA]\+)[-!=~^]*/
 
   (* View: mode
 "-", or 3-4 bytes. Optionally starts with a "~". *)
--- a/lenses/tmpfiles.aug
+++ b/lenses/tmpfiles.aug
@@ -58,8 +58,8 @@ Not all letters are valid.
   let type     = /([fFwdDevqQpLcbCxXrRzZtThHaAm]|[fFwpLcbaA]\+)[-!=~^]*/
 
   (* View: mode
-"-", or 3-4 bytes. Optionally starts with a "~". *)
-  let mode     = /(-|~?[0-7]{3,4})/
+"-", or 3-4 bytes. Optionally starts with a "~" or a ":". *)
+  let mode     = /(-|(~|:)?[0-7]{3,4})/
 
   (* View: age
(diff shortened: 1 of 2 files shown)


georgehansper:
#795  mentions the `^` and `~` notation in the systemd source

I can't see any mention of the `:` notation for the mode or the man page for tmpfiles.d

Can you please let me know where this new form is coming from?

ptoscano:
> I can't see any mention of the `:` notation for the mode or the man page for tmpfiles.d
> 
> Can you please let me know where this new form is coming from?

It is documented in the man page:
https://github.com/systemd/systemd/blob/25e12db5e4c0edf29c4d020b825a8d667a3aa8a3/man/tmpfiles.d.xml#L550-L552

lenses/tmpfiles.aug
```diff
@@ -50,15 +50,16 @@ Empty lines *)
...
   (* View: type
 One letter. Some of them can have a "+" and all can have an
-exclamation mark ("!") and/or minus sign ("-").
+exclamation mark ("!"), a minus sign ("-"), an equal sign ("="),
+a tilde character ("~") and/or a caret ("^").
 
 Not all letters are valid.
 *)
-  let type     = /([fFwdDevqQpLcbCxXrRzZtThHaAm]|[fFwpLcbaA]\+)!?-?/
+  let type     = /([fFwdDevqQpLcbCxXrRzZtThHaAm]|[fFwpLcbaA]\+)!?-?=?~?\^?/
```
georgehansper, line 58: The expression
  `!?-?=?~?\^?`
makes each of these characters optional on its own, but when used in combination, the expression implies a specific order that is not mentioned in the documentation.
eg: the expression would accept `~^` but not `^~`

It would be better to use a more loose expression like this:
 `[-!=~^]*`

I'm not sure if this example would need parenthesis `([-!=~^]*)`

lenses/tmpfiles.aug
```diff
@@ -50,15 +50,16 @@ Empty lines *)
...
 
   (* View: mode
-"-", or 3-4 bytes. Optionally starts with a "~". *)
-  let mode     = /(-|~?[0-7]{3,4})/
+"-", or 3-4 bytes. Optionally starts with a "~" or a ":". *)
+  let mode     = /(-|(~|:)?[0-7]{3,4})/
```
georgehansper, line 62: My reading of the documentation is that ~ and : would not be used together, so this is OK
~ applies to existing files (only), while : applies to new files (only)

georgehansper (changes requested):

lenses/tmpfiles.aug
ptoscano: makes sense, even the documentation says that there can be more than one -- changed, thanks!

georgehansper (approved):

lersek:
@ptoscano thank you very much Pino!

PR #809, opened 2023-04-11
TOML: support trailing commas

### LINKED ISSUE #753: array with trailing comma in Toml lens
Toml lenses won't load file containing array with trailing comma like this: `cors_allowed_methods = ["HEAD", "GET", "POST", ]`. As [described](https://toml.io/en/v1.0.0#array) in language spec (look for string `2, # this is ok`) this should be ok.

qubidt:
Add support for trailing commas in TOML arrays. Fixes #753.


georgehansper (approved):
I was having a look at this, and there are further improvements that could be made to the handling of TOML files.

This is good start, though

PR #810, opened 2023-04-19
Fix bug from PR#691 where the nodes of a newly created file are lost upon a subsequent load operation

georgehansper:
Pull Request #691 introduced a new function to path expressions `modified()`
This function exposed the value of the internal flag `dirty` on a node.
In order for `modifed()` to reflect _only_ the changed nodes, some of the logic around the `dirty` flag was modified.
This caused a bug which appeared after a `save`operation, where upon a subsequent `load` operation, where the nodes of a newly created file were inadvertently  removed.
This PR fixes the inadvertent cleanup on the subsequent load operation, if the file has been saved.
Unsaved nodes under `/files` are still cleaned up by the load operation

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -1516,6 +1516,10 @@ static int tree_save(struct augeas *aug, struct tree *tree,
                 }
             }
             if (transform != NULL) {
+                /* If this file did not previously exist and is being created by augeas,
+                 * then the 'file' flag in the node t will not be set yet. Set it now
+                 */
+                t->file = true;
                 int r = transform_save(aug, transform, tpath, t);
                 if (r == -1)
                     result = -1;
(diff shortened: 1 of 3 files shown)


ekohl (commented):
In https://github.com/voxpupuli/puppet-augeasproviders_core/pull/49#issuecomment-1522394658 there's a verification that this solves the bug showing up in Puppet so :+1: for this.

georgehansper:
Thanks for confirming this @ekohl

ekohl:
Thanks for this. Any plans for a release that includes it? While we do have a workaround in our code now (https://github.com/voxpupuli/puppet-augeasproviders_core/commit/da34348698fad3a280e496f776f836a385b125db), it would be great to have the real fix in place. Especially since it can take some time before a new release makes it to all platforms.

georgehansper:
I'd like to do a minor release soon

georgehansper:
augeas-1.14.1 is now released, with this fix included

PR #815, opened 2023-07-19
read subdirectories from /etc/zabbix dir

smgrol:
PR for read configuration files in subdirectories in /etc/zabbix directory with augeas


georgehansper:
The `**`  notation has no particular significance in Augeas

Can you please change this to a single `*` (for clarity/consistancy) ?

smgrol:
hi!
thanks for review

changed to single *

georgehansper (approved):
Thanks!

Issue #817, opened 2023-09-27
Implicit function delcaration error with Apple Clang


p-linnane:
Hello 👋 . I'm a maintainer for the [Homebrew](https://brew.sh) project. We've noticed the following error with Apple Clang when trying to compile augeas. We are currently working around this by passing `-Wno-implicit-function-declaration`, but hoping to get an upstream fix. Thanks!

```
lens.c:979:9: error: call to undeclared function 'dump_lens_tree'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        dump_lens_tree(lens);
        ^
1 error generated.
```

igalic:
this code is a bit weird. both `dump_lens_tree` and the `debugging` function guarding its call are gated by `# if ENABLE_DEBUG`. but unlike `debugging`, there's no `# else` branch declaring a (useless) `dump_lens_tree` version to appease the compiler.

so, despite the fact that the compiler sees this:

```c
if ((0))
      dump_lens_tree(lens);
```

it doesn't just delete the code from the AST, it tries to actually analyse it.

A possibly fix would be to declare the two dump functions regardless of whether Debugging is enabled. Ideally we could declare them `static`, so as to not pollute the exported functions

georgehansper:
Thanks for pointing this out @igalic

The function `dump_lens_tree()` is defined within an `#if ENABLE_DEBUG` ... `#endif` clause, as is the corresponding function declaration in the header file `lens.h`

I was wondering why the compiler did not complain that the function didn't exist at all after complaining that it was an "undeclared function".

I expect it is because the of the `if(0)`, which allows the compiler to optimise it out of existance in the code.

The "undeclared function" warning probably happens on a eariler pass of the compiler over the code, before it does the optimisation

PR#818 puts the code inside an `#if ENABLE_DEBUG` clause.

As an alternative, the `lens.h` could be updated along the lines of `debugging()` in `internal.h`, like this:

```
#if ENABLE_DEBUG
void dump_lens_tree(struct lens *lens);
void dump_lens(FILE *out, struct lens *lens);
#else
#define dump_lens_tree(lens) (void)0
#endif
```

I'm thinking that this is closer to the existing solution applied for the function `debugging()` which is in a similar situation

The net result is that when `ENABLE_DEBUG` is false, to compiler sees the dummy macros in place of the functions and stops complaining about them being "undeclared functions"

p-linnane:
[...]

PR #818, opened 2023-09-30
Fix some compiler warnings in lens.c, augmatch.c

georgehansper:
Tested with:

```
./autogen.sh --enable-debug=no --enable-compile-warnings=maximum
make distcheck
```

and

```
./autogen.sh --enable-debug=yes --enable-compile-warnings=maximum
make distcheck
```

Fixes issue #817

### DIFF
--- a/src/augmatch.c
+++ b/src/augmatch.c
@@ -302,7 +302,7 @@ static char *guess_lens_name(const char *file) {
 
 int main(int argc, char **argv) {
     int opt;
-    cleanup(aug_closep) struct augeas *aug;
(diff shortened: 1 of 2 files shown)


georgehansper:
I've upated this Pull Request along lines of the discussion in #817
ie. to use `#define dump_lens_tree(...` instead of `#if ENABLE_DEBUG...`

The end-result is the same (the compiler warning is gone), but it "fits" better with the existing code and is slightly more readable

igalic (commented):
👍

PR #821, opened 2023-11-25
Fix lens ntp.aug (tos/minclock) and support ntpsec conf file (Debian12)

FrancoisRegisDegott-eaton:
- fixed lens parser fails on 'tos/minclock' (/etc/ntpsec/ntp.conf on debian12 with that line "tos minclock 4 minsane 3" )
- filter : add /etc/ntpsec/ntp.conf

### DIFF
--- a/lenses/ntp.aug
+++ b/lenses/ntp.aug
@@ -117,8 +117,7 @@ module Ntp =
             orphan stratum | orphanwait delay] *)
 
     let tos =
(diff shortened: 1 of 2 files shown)


igalic (commented):
can you please extend https://github.com/hercules-team/augeas/blob/master/lenses/tests/test_ntp.aug ?

igalic (commented):
👍

PR #825, opened 2024-02-27
bootstrap and autogen: add the --no-git option

shal4:
Some users already have gnulib downloaded. Added the --no-git option to allow users to skip re-downloading gnulib.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -13,6 +13,8 @@ Options:
                           have gnulib sources on your machine, and
                           do not want to waste your bandwidth downloading
                           them again.
+ --no-git                 Do not use git to update gnulib. Requires that
+                          --gnulib-srcdir point to a correct gnulib snapshot
  --help                   Print this message
  any other option         Pass to the 'configure' script verbatim
 
@@ -30,6 +32,8 @@ test -z "$srcdir" && srcdir=.
 THEDIR=`pwd`
 cd $srcdir
 
+no_git=
+
 # Split out options for bootstrap and for configure
 declare -a CF_ARGS
 for option
@@ -40,6 +44,8 @@ do
     exit;;
   --gnulib-srcdir=*)
     GNULIB_SRCDIR=$option;;
+  --no-git)
+    no_git=" --no-git";;
   *)
     CF_ARGS[${#CF_ARGS[@]}]=$option;;
   esac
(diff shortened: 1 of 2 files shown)


georgehansper:
I've been looking over the bootstrap script, and how the --gnulib-srcdir option works

"gnulib" has been made a submodule of the augeas git-repo

When the main repository (augeas) is cloned, the submodules (gnulib) are not included.
These need to be downloaded seperately, and a checkout operation done

This is a 2 step process:

* git submodule init
		...copies the URL of gnulib from .gitmodules to .git/config

* git submodule update
		...clones the URL to .git/modules/gnulib as a bare repo
		...does a checkout of this repo to .gnulib at the commit shown by "git submodule status"

This "locks" the gnulib version to a specific commit from the gnulib repo

The --gnulib-srcdir option inserts a step in between init and update, changing the URL in .git/config to the one given.
The updated URL is used to clone into .git/modules/gnulib

Like yourself, I have noticed that gnulib is available as a "package" on various systems.
However the current implementation of "bootstrap" is dependant on it being able to checkout a specific commit from the gnulib git-repository

Personally, I think that the argument name --gnulib-srcdir is a bit misleading, and something like --gnulib-gitdir would be more descriptive

Additionally, simply specifying a valid, local gnulib git-repository is not enough, as this fails with a message:

```
  fatal: transport 'file' not allowed
```

To fix this, the users global gitconfig (ie $HOME/.gitconfig) must be updated to contain:

```
[protocol "file"]
	allow = always
```

Probably any valid git URL would suffice, except that "bootstrap" looks for a subdirectory .git of the specified argument, which must be a directory

In short, the argument --gnulib-srcdir is itself quite complex, and not well documented by the brief help message

I do have one reservation about your approach here, and that is the argument --no-git subject to broader interpretation

With all the above in mind, and a strong desire to maintain backwards-compatability, can I suggest that:

a) the argument --no-git be instead replaced by --gnulib-nogit (or if you prefer --gnulib-notgit or --gnulib-no-git)

b) instead of being a "flag" the argument --gnulib-nogit is a complete replacement for --gnulib-srcdir, and takes a directory as a value

c) if --gnulib-nogit=/path/to/gnulib is specified, the user is given a warning that their version of gnulib has not been tested against the current release of augeas

d) the behaviour of --gnulib-srcdir be updated to warn the user if:
			the subdirectory .git/modules/gnulib exists, which results in it being ignored
			"git config --global protocol.file.allow" != always ie. .gitconfig must contain the following, 
			which is required to allow the directory to be cloned as a submodule

```
				[protocol "file"]
					allow = always
```

e) (optionally) an new argument --gnulib-gitdir be created as an alias for --gnulib-srcdir

f) the --help message be updated to cover all of the above nuances

I am happy to assist with any of the above

PR #838, opened 2024-07-17
lenses/fstab.aug: Allow comma after the last option

### LINKED ISSUE #832: cannot parse /etc/fstab with trailing "," in the fs_mntops field
```
$ mkdir -p root/etc
$ echo '/dev/mapper/foo-bar / xfs defaults, 0 0' > root/etc/fstab
$ cat root/etc/fstab
/dev/mapper/foo-bar / xfs defaults, 0 0
```
Note extra "," in the fourth field.  Apparently the tools which normally parse fstab are fine with that.

```
$ augtool -r $PWD/root
augtool> get /augeas/files/etc/fstab/error/message
/augeas/files/etc/fstab/error/message = Iterated lens matched less than it should
augtool> get /augeas/files/etc/fstab/error/line
/augeas/files/etc/fstab/error/line = 1
augtool> get /augeas/files/etc/fstab/error/char
/augeas/files/etc/fstab/error/char = 34
```

Augeas fails to parse the file at the extra "," character.

tupyy:
This PR adds a new lens that allows a comma after the last option:
```
/dev/mapper/foo-bar / xfs defaults, 0 0
```

Fixes: https://github.com/hercules-team/augeas/issues/832

Signed-off-by: cosmin@redhat.com

### DIFF
--- a/lenses/fstab.aug
+++ b/lenses/fstab.aug
@@ -5,6 +5,7 @@ module Fstab =
 
   let sep_tab = Sep.tab
   let sep_spc = Sep.space
+  let sep_comma_tab = del /,?[ \t]+/ "\t"
   let comma   = Sep.comma
   let eol     = Util.eol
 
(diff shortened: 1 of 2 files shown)


rwmjones:
Please squash the commits into one (you'll have to force push to your branch).

Also needs a `Signed-off-by` in the commit message.

Also remove `fstab -` prefix and either replace it with nothing or with `fstab:`.  See existing commit messages in the project (which are not very consistent, but nothing uses `-`).

tupyy:
@rwmjones done

rwmjones:
You need to add the `Fixes:`.  Also git commit messages are plain text, so using back ticks is not necessary.

rwmjones:
Also it should be `Signed-off-by`, see https://www.kernel.org/doc/html/latest/process/submitting-patches.html

tupyy:
ptal

rwmjones:
This looks good now.  I also applied it to my local tree and ran the tests and those pass.

Issue #839, opened 2024-07-31
Augeas can't parse systemd unit file entry `ExecStartPre=+-command`


rwmjones:
With the latest SSSD 2.10 was added a systemd unit file that contains:
```
ExecStartPre=+-/bin/chown -f sssd:sssd /etc/sssd/sssd.conf
ExecStartPre=+-/bin/chown -f -R sssd:sssd /etc/sssd/conf.d
ExecStartPre=+-/bin/chown -f -R sssd:sssd /etc/sssd/pki
ExecStart=/usr/sbin/sssd -i ${DEBUG_LOGGER}
```
The new `ExecStartPre=+-...` entries were added in this version and Augeas fails to parse them:
```
$ augtool print /augeas//error | grep sssd.service
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error = "parse_failed"
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error/pos = "413"
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error/line = "14"
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error/char = "0"
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error/lens = "/usr/share/augeas/lenses/dist/inifile.aug:497.25-.43:"
/augeas/files/etc/systemd/system/multi-user.target.wants/sssd.service/error/message = "Get did not match entire input"
/augeas/files/lib/systemd/system/sssd.service/error = "parse_failed"
/augeas/files/lib/systemd/system/sssd.service/error/pos = "413"
/augeas/files/lib/systemd/system/sssd.service/error/line = "14"
/augeas/files/lib/systemd/system/sssd.service/error/char = "0"
/augeas/files/lib/systemd/system/sssd.service/error/lens = "/usr/share/augeas/lenses/dist/inifile.aug:497.25-.43:"
/augeas/files/lib/systemd/system/sssd.service/error/message = "Get did not match entire input"
```

I found through experimentation that the actual problem is the use of `+` and `-` characters together.  [This is valid for systemd](https://www.man7.org/linux/man-pages/man5/systemd.service.5.html#COMMAND_LINES), but Augeas only understands a single character here (eg. just `+` or just `-` works, but not two or more).

rwmjones:
FYI this is the downstream bug filed against RHEL: https://issues.redhat.com/browse/RHEL-50778

[...]

PR #841, opened 2024-08-06
lenses/systemd.aug: Allow "+"(fullprivileges) command flag

tupyy:
This commit allows the "+"(fullprivileges) command flag along with "-" and "@" flags.

Fixes: #839

### DIFF
--- a/lenses/systemd.aug
+++ b/lenses/systemd.aug
@@ -72,8 +72,8 @@ let sto_value = store /[^# \t\n]*[^# \t\n\\]/
 let value_sep = del /[ \t]+|[ \t]*\\\\[ \t]*\n[ \t]*/ " "
 
 (* Variable: value_cmd_re
-   Don't parse @ and - prefix flags *)
-let value_cmd_re = /[^#@ \t\n\\-][^#@ \t\n\\-][^# \t\n\\]*/
+   Don't parse @, - and + prefix flags *)
+let value_cmd_re = /[^-#@+ \t\n\\][^# \t\n\\]*/
 
 (* Variable: env_key *)
 let env_key = /[A-Za-z0-9_]+(\[[0-9]+\])?/
@@ -111,11 +111,12 @@ let entry_multi  = entry_fn entry_multi_kw
                        . Build.opt_list entry_value value_sep )?
 
 (* View: entry_command_flags
-   Exec* flags "@" and "-".  Order is important, see systemd.service(8) *)
+   Exec* flags "@", "-" and "+".  Order is *not* important *)
 let entry_command_flags =
-     let exit  = [ label "ignoreexit" . Util.del_str "-" ]
+     let fullprivileges = [ label "fullprivileges" . Util.del_str "+" ]
+  in let exit  = [ label "ignoreexit" . Util.del_str "-" ]
   in let arg0  = [ label "arg0" . Util.del_str "@" ]
-  in exit? . arg0?
+  in (fullprivileges | exit  | arg0)*
 
 (* View: entry_command
    Entry that takes a command, arguments and the optional prefix flags *)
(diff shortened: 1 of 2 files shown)


rwmjones:
For S-o-b you should use your full name and email, like `Signed-off-by: Cosmin Tupangiu <cosmin@redhat.com>` as this is a [pseudo-legal declaration](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for#1962112).

The reporter wasn't me, it was `Yongkui Guo <yoguo@redhat.com>`

When applying the patch I see trailing whitespace, please fix as well:
```
Applying: Allow fullprivileges command flag
.git/rebase-apply/patch:29: trailing whitespace.
     let fullprivileges = [ label "fullprivileges" . Util.del_str "+" ] 
warning: 1 line adds whitespace errors.
```

I tested it and the tests passed for me and it otherwise looks good.

tupyy:
@rwmjones done

rwmjones:
LGTM now

tupyy:
> @tupyy Could you please rebase this on top of the current head.

@rwmjones done

georgehansper:
This issue is a bit bigger than it seems

The problem described in issue #839 is correct, and the fix in Pull Request #841 fixes the issue described

However, there a 3 issues which are not addressed, and probably should be

* flag order
  systemd does not consider the order of the flags significant, but this lens does
  This is a characteristic of the existing lens. ie. This problem not introduced bt this change, but it is perpetuated by it
  eg while this lens will read and generate a line

```
    ExecStart=+-/usr/sbin/sshd -D $OPTIONS
```

this lens will neither read nor generate a line with the flags in the reverse order

```
    ExecStart=-+/usr/sbin/sshd -D $OPTIONS`
```

* unhandled flags
  There are other flags that could be considered - '!' '!!' and ':'
  Note that the systemd man page states that only one of '+' '!' and '!!' may be used.
  While it might be nice to have this restriction taken into account, it is not up to augeas to enforce the syntax rules of the target software, and it would probably contribute significantly to the complexity of the lens

* backwards compatability
  Using the existing lens, some people may be using statements such as
```
    set /files/lib/systemd/system/sshd.service/Service/ExecStart/command "+/usr/sbin/sshd"
```
  I'm not sure if it's possible, but it would be nice if the updated lens would still accept the old set statement
  I do notice that the existing lens does not accept a value of "-/usr/sbin/sshd" so this may not be possible to achieve

tupyy:
Hello,
As already stated, enforcing systemd cmd flags rules in Augeus, in my opinion, is not very helpful. It should be systemd that check and enforce these flags.
That being said, I don't think it's possible to write lens for the flags that check and enforce the rules. Currently, every flag it's a lens that complicates the writing of `+-` or `-+`. At least, I don't know how to write them. I've tried but the problem is Augues complains that the multiple lenses matched the flag. 

The easiest way is to bundle everything together like `let cmd_flags = /[+-!@:]+/` and let systemd handle the rules.
But this, I think it will break the backwards compatibility.

georgehansper:
Hi Cosmin,

I think we are agreed that Augeas should not try to enforce systemd flag rules. I only mentioned it to say not to waste effort on this.

I like the idea of having cmd_flags and an explicit value (as you have shown), but as you mention, we should try to keep the lens backwards compatible.

Allowing an arbitrary flag order can be achieved by changing the following 2 lines:
 
```
-let value_cmd_re = /[^#@+ \t\n\\-][^#@+ \t\n\\-][^# \t\n\\]*/
+let value_cmd_re = /[^-#@+ \t\n\\][^# \t\n\\]*/
```
and
```
-  in fullprivileges? . exit? . arg0?
+  in (fullprivileges | exit |  arg0 )*
```

Notice that by restricting just the 1st character of `value_cmd_re` to being a non-flag character, this creates an boundary between the flags and the command. Augeas can use this to unambiguously distinguish the start of the command

tupyy:
Hi George,

Thank you for the suggestions. I've updated the PR accordingly.
PTAL

Cosmin

PR #842, opened 2024-08-06
Fix gcc warnings

rwmjones:
Fix miscellaneous GCC warnings.  In particular this causes the compile to fail when using `configure --enable-compile-warnings=error` and using a recent version of GCC.


tupyy:
LGTM

georgehansper:
Thanks for these fixes, @rwmjones 
I completely missed the fact the the args to reallocarray() in augprint.c were back-to-front

Issue #852, opened 2025-03-10
NULL Pointer Dereference in re_case_expand


sae-as-me:
## Affected Projects
augeas 1.14.1 ([https://github.com/hercules-team/augeas](https://github.com/hercules-team/augeas))
## Problem Type
CWE-476: NULL Pointer Dereference
### Details
The vulnerability arises in the `re_case_expand` function defined in `src/fa.c` at line `4391`. Function `fa_expand_nocase` lacks a check for whether the pointer is null when calling function `re_case_expand`, resulting in null pointer dereference and causing the program to report segmentation fault.
```c++
int fa_expand_nocase(const char *regexp, size_t regexp_len,
                     char **newregexp, size_t *newregexp_len) {
    int result, r;
    struct re *re = NULL;
    struct re_parse parse;
    struct re_str str;

    *newregexp = NULL;
    MEMZERO(&parse, 1);
    parse.rx = regexp;
    parse.rend = regexp + regexp_len;
    parse.error = REG_NOERROR;
    re = parse_regexp(&parse);
    if (parse.error != REG_NOERROR)
        return parse.error;

    r = re_case_expand(re);    // re may be a null pointer
    //...
}
```

![Image](https://github.com/user-attachments/assets/d4149a6b-aec8-4ab0-9919-4a56d0a5a897)

And resulting in null pointer dereference in `re_case_expand` when access `re->type`.
```c++
static int re_case_expand(struct re *re) {
    int result = 0, r1, r2;

    switch(re->type) {    //null pointer dereference
    //case ...
    default:
        assert(0);
        abort();
        break;
    }
    return result;
}
```

![Image](https://github.com/user-attachments/assets/f00ecd83-394b-4184-9cd0-b60bebe4b0d3)

### PoC
**Steps to reproduce:**

1. Clone the Augeas repository and build it with the OSS-Fuzz configuration.
- ENV
```sh
export CC='clang'
export CXX='clang++'
export CFLAGS='-fsanitize=address -g'
export CXXFLAGS='-fsanitize=address -g'
```
- harness from OSS-Fuzz
```c++
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include <stdlib.h>

// Augea includes
#include "augeas.h"
#include "config.h"
#include "fa.h"
#include "internal.h"


/*
 * FA fuzzer. 
 */
extern "C"
int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size){
	if(Size<3){
		return 0;
[...]

PR #854, opened 2025-03-24
CVE-2025-2588: return _REG_ENOSYS if no specific error was set yet pa…

abbra:
…rse_regexp failed

parse_regexp() supposed to set an error on the parser state in case of a failure. If no specific error was set, return _REG_ENOSYS to indicate a generic failure.

Fixes: https://github.com/hercules-team/augeas/issues/671
Fixes: https://github.com/hercules-team/augeas/issues/778
Fixes: https://github.com/hercules-team/augeas/issues/852

### DIFF
--- a/src/fa.c
+++ b/src/fa.c
@@ -3550,6 +3550,8 @@ static struct re *parse_regexp(struct re_parse *parse) {
     return re;
 
  error:
+    if (re == NULL && parse->error == REG_NOERROR)
+        parse->error = _REG_ENOSYS;
     re_unref(re);
     return NULL;
 }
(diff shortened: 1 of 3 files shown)


rwmjones:
Added to our downstream Fedora branch: https://github.com/rwmjones/augeas/tree/fedora-43

src/fa.c
```diff
@@ -32,6 +32,7 @@
 #include <config.h>
 #include <limits.h>
 #include <ctype.h>
+#include <regex.h>
```
georgehansper, line 35: This include is not required, as it is included within file "fa.h" below

georgehansper:
Other than the above, this PR looks fine
Thanks for adding the test-case, too

src/fa.c
abbra: Removed, thank you.

georgehansper (approved):

PR #859, opened 2025-04-02
Fix preview when spans are enabled

amousset:
The `preview` function does not work as expected when the `AUG_ENABLE_SPAN` flag is passed. This path computed by `tree_source` seems incorrect in this case:

* When the spans are not enabled, the `tree_source` call in the test returns `/files/etc/hosts`.
* When the spans are enabled, the `tree_source` call in the test returns `/home/.../augeas/tests/root/etc/hosts` which then makes the preview fail.

The added tests fails, and is fixed by removing the special case when span are defined in `tree_source`, but I don't grasp the possible implications of this change.

### DIFF
--- a/src/augeas.c
+++ b/src/augeas.c
@@ -1933,14 +1933,7 @@ tree_source(const augeas *aug, struct tree *tree) {
         tree = tree->parent;
 
     if (tree->file) {
-        if (tree->span == NULL) {
-            int r;
-            r = ALLOC(tree->span);
-            ERR_NOMEM(r < 0, aug);
-            tree->span->filename = make_string(path_of_tree(tree));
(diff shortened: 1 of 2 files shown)


georgehansper:
Thank you for reporting this problem, and also my appologies for the slow response

Like you, I have no idea why the behaviour of tree_source() differs when spans are enabled.

aug_preview() is specifically trying to find the 'file' node for the given path, and tree_source() seemed like an easy way to get this
As it turned out, tree_source() was not the right choice for this

I do not think it is a good idea to change the behaviour of tree_source() to suit the newer aug_preview() function
I think this should be fixed within aug_preview() and leave tree_source() alone

I do like the additional test included in the PR, so I would appreciate it if you could update this PR to just include the changes to tests/test-api.c

PR #861, opened 2025-04-15
lenses/iptables.aug: Support ipset module

### LINKED ISSUE #860: lense for iptables does not natively support second argument required for set module (match-set)
Not to confuse set command in augeas, but looking to support adding rules in iptables for ipset.
Example of the target rule
``` -I INPUT -m set --match-set test src -j DROP```

To test this I'd started with creating a rule through augtool. However its not going to allow for passing a third argument in the set command. 
```
defnode rule $filter/insert[.='INPUT'] "INPUT"
set $rule/match "set"
set $rule/match-set "test" "src"
set $rule/jump "DROP"
```
From there added the rule directly to Iptables and tried to print the rule which returns an error on match
Without any modification to the iptables lens this will return the following error
```
Error in /etc/sysconfig/iptables:144.0 (parse_failed)
  Iterated lens matched less than it should
  Lens: /usr/share/augeas/lenses/dist/iptables.aug:84.10-.32:
    Last matched: /usr/share/augeas/lenses/dist/iptables.aug:81.16-.43:
    Next (no match): /usr/share/augeas/lenses/dist/util.aug:22.27-.34:
```
If I am interpreting lens correctly  --match-set will match ipt_match however its syntax is similar is similar to [tcp-flags](https://github.com/hercules-team/augeas/blob/master/lenses/iptables.aug#L42-L48) which then I could clone the binding/definition. Is there an existing/alternate solution to this that I am missing?

### PR DESCRIPTION
**Pull request by citizenbilly**
Iptables includes a module for integration with ipset.
However, augeas does not appear to support the second argument required for the iptables module: ```---match-set```.

Example Iptables rule:
```-I INPUT -m set --match-set <setname> src -j DROP```

Example of adding the rule:
```
defnode rule \$filter/insert[.='INPUT'][last()+1] "INPUT"
set \$rule/match "set"
set \$rule/ipset_flags "<setname>"
set \$rule/ipset_flags/set "src"
set \$rule/jump "DROP"
```

Fixes: [hercules-team#860](https://github.com/hercules-team/augeas/issues/860)

### DIFF
--- a/lenses/iptables.aug
+++ b/lenses/iptables.aug
@@ -47,6 +47,15 @@ let tcp_flags =
       spc . dels "--tcp-flags" .
       spc . flag_list "mask" . spc . flag_list "set" ]
 
+let ipset_flags =
+  let flags = /src|dst/ in
+  let match_flag (name:string) =
+    Build.opt_list [label name . store flags] (dels ",") in
+  [ label "ipset_flags" .
+      spc . dels "--match-set" .
+      spc . store /[a-zA-Z-][a-zA-Z0-9-]+/ .
+      spc . match_flag "set" ]
+
 (* misses --set-counters *)
 let ipt_match =
   let any_key = /[a-zA-Z-][a-zA-Z0-9-]+/ -
@@ -65,6 +74,7 @@ let ipt_match =
     |neg_param "fragment" "f"
     |param "match" "m"
     |tcp_flags
+    |ipset_flags
     |any_param)*
 
 let chain_action (n:string) (o:string) =

### CODE REVIEW

**Comment by georgehansper** [comment]
The changes here work as expected
Can I suggest that the labels "ipset_flags" and "set" be changed to reflect the intent more closely
For the example given, the resulting augeas tree is

```
/files/etc/sysconfig/iptables/table[1]/insert = "INPUT"
/files/etc/sysconfig/iptables/table[1]/insert/match = "set"
/files/etc/sysconfig/iptables/table[1]/insert/ipset_flags = "setname"
/files/etc/sysconfig/iptables/table[1]/insert/ipset_flags/set = "src"
/files/etc/sysconfig/iptables/table[1]/insert/jump = "DROP"
```

I think it would be better to use:

*  "ipset" in place of "ipset_flags"
*  "flags" in place of "set"

The term "flags" here may seem a little too general, but this is the term used in the ipset man page

The resulting (updated) tree for the input

```
-A INPUT -m set --match-set setname src,dst -j ACCEPT
```

```
/files/etc/sysconfig/iptables/table[1]/append = "INPUT"
/files/etc/sysconfig/iptables/table[1]/append/match = "set"
/files/etc/sysconfig/iptables/table[1]/append/ipset = "setname"
/files/etc/sysconfig/iptables/table[1]/append/ipset/flags[1] = "src"
/files/etc/sysconfig/iptables/table[1]/append/ipset/flags[2] = "dst"
/files/etc/sysconfig/iptables/table[1]/append/jump = "ACCEPT"
```

Can you please update this pull-request as outlined above?

Issue #846, opened 2024-10-16
Augeas 0.13.0 and 0.14.0 barf with rsyslog8 if statements.


xbeaudouin:
Hello,

We have found some specific issues with augeas and rsyslog8 ìf`statements.
This has been discovered on FreeBSD and we have been able to reproduce it on Ubuntu 22.04 with augeas 0.13.0-1, and on FreeBSD on augeas 0.14.0.

Create a `/tmp/etc/rsyslogd.conf` with following contents:

```
# Consult the rsyslog.conf(5) manpage, and the comprehensive on-line
# documentation at
# https://www.rsyslog.com/doc/v8-stable/configuration/index.html
#
# Derived from
# https://cgit.freebsd.org/src/tree/usr.sbin/syslogd/syslog.conf

module(load="immark")   # provides --MARK-- message capability
module(load="imuxsock") # provides support for local system logging
module(load="imklog")   # kernel logging

*.err;kern.warning;auth.notice;mail.crit                /dev/console
*.notice;authpriv.none;kern.debug;lpr.info;mail.crit;news.err   /var/log/message
s
security.*                      /var/log/security
auth.info;authpriv.info         /var/log/auth.log
mail.info                       /var/log/maillog
cron.*                          /var/log/cron

if $programname != "devd" then {
   *.=debug                     /var/log/debug.log
   *.emerg                      action(type="omusrmsg" users="*")
   daemon.info                  /var/log/daemon.log
}
```

Then just execute:
```
# augtool -r /tmp errors
Error in /etc/rsyslog.conf:20.0 (parse_failed)
  Get did not match entire input
  Lens: /usr/local/share/augeas/lenses/dist/rsyslog.aug:92.10-.43:
```

Remove the lines starting with `if ...`, then `augtool` will work perfectly.

georgehansper:
Thank you for flagging this issue
It will require a update to the lens

xbeaudouin:
I have pushed and update of the lense. This update make augeas working at it should works when parsing rsyslogd.conf.
Kind regards

xbeaudouin:
ping ?

xbeaudouin:
Hello, 
Any news about this lense update?

PR #863, opened 2025-06-02
lenses/rsyslog.aug: Fix if xxx then { }

xbeaudouin:
If the rsyslog.conf have :
  if $programname != "devd" then {
  }

The lense fails with a parse_failed.

This fix bug #846


xbeaudouin:
ping ?

xbeaudouin:
I have cleaned some carriage returns that make the code change difficult to understand. So if someone can review the code, and the idea.

xbeaudouin:
Ping ?

xbeaudouin:
ping ?

PR #875, opened 2026-01-12
lenses/iproute2: Ignore Readme, add new directory

twied-redhat:
From v6.5.0, iproute2 supports stateless configuration pattern. This means that iproute2 now reads its config from /etc/iproute2/FOO, and, if it does not exist, falls back to /usr/share/iproute2/FOO.

A "README" file in /etc/iproute2 explains the matter, but the file itself needs to be excluded from the lens.

### DIFF
--- a/lenses/iproute2.aug
+++ b/lenses/iproute2.aug
@@ -7,4 +7,9 @@ module IPRoute2 =
 
   let lns = ( empty | Util.comment | record ) *
 
-  let xfm = transform lns (incl "/etc/iproute2/*" . Util.stdexcl)
+  let filter = incl "/etc/iproute2/*"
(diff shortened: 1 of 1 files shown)


georgehansper:
I notice that some distros (almalinux 9, almalinux 10) also have a README file in /usr/share/iproute2

Can we please exclude this file, too, either by adding the full path as
```
  . excl "/usr/share/iproute2/README"
```

or just the filename, regardless of the full pathname (replaces the excl line in this PR)

```
 . excl "README"
```

georgehansper (approved):
OK, other than the omission of
```
 . excl "/usr/share/iproute2/README"
```

PR #883, opened 2026-06-09
  ./configure   make && make installRename COPYING to Home

romeroyair216-ops:


lutter:
Since `COPYING` is the proper place for the license, we will not move it to a different filename

PR #884, opened 2026-06-09
fix out-of-bounds read in regexp parser peek()

rootvector2:
`peek` derefs `*parse->rx` without a `more()` check, so a length-delimited regexp that ends on a peek (`fa_compile("a|", 2)`, `"("`) reads one byte past `parse->rend`; found auditing the parser's length handling, fixed by guarding `peek` like `match`/`next` already do so it respects the buffer length instead of assuming NUL termination.


rootvector2:
gentle ping

PR #886, opened 2026-06-15
cast to unsigned char before ctype calls

rootvector2:
noticed in the path-expression lexer: it, the regex example builder and the error formatter pass a plain `char` to `isspace`/`isalpha`/`isprint`, so a high byte like a utf-8 path label reaches `ctype` as a negative `int` (undefined behavior, an out-of-bounds table read where the ctype table has no negative slots); cast each argument to `unsigned char`.

### DIFF
--- a/src/augrun.c
+++ b/src/augrun.c
@@ -126,7 +126,7 @@ static char *nexttoken(struct command *cmd, char **line, bool path) {
 
     s = *line;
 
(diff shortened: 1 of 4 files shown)


rootvector2:
gentle ping

lutter (approved):

lutter:
Sorry for the delay, and thanks a ton for spotting and fixing this!

PR #893, opened 2026-08-16
Simplevars: take over /etc/adduser.conf from Shellvars

zakinko:
`/etc/adduser.conf` is in the `Shellvars` filter. Of the three systems that ship
the file, only one treats it as a shell script.

**OpenBSD** — `adduser(8)` is perl and evals each line
(`usr.sbin/adduser/adduser.perl`, `config_read`), so the file it writes is perl:

```
verbose = 1
defaultpasswd = "yes"
path = ('/bin', '/usr/bin', '/usr/local/bin')
```

`Shellvars` cannot parse that, so on a stock install the file is simply
unavailable:

```
/augeas/files/etc/adduser.conf/error = "parse_failed"
/augeas/files/etc/adduser.conf/error/line = "41"
/augeas/files/etc/adduser.conf/error/lens = ".../shellvars.aug:251.12-.60:"
```

**Debian** — `adduser` is perl as well, and reads the file with

```perl
m/^\s*([_a-zA-Z0-9]+)\s*=\s*(...)/
```

Whitespace around `=` is accepted and surrounding quotes are stripped.
`Shellvars` does not reject the spaced form; it reads it as a command, so the
setting silently disappears from `/files/etc/adduser.conf/DSHELL`:

```
DSHELL = /bin/bash

  { "@command" = "DSHELL"
    { "@arg" = "= /bin/bash" } }
```

**FreeBSD** — here it really is shell. `adduser(8)` is `/bin/sh` and sources it
(`. "${ADDUSERCONF}"`).

**NetBSD** has no `adduser(8)` at all; `useradd(8)` reads `/etc/usermgmt.conf`
instead, so it is unaffected.

`Simplevars` is the grammar Debian's parser accepts, it keeps the perl form as
an opaque value, and it builds the same tree as `Shellvars` for everything the
three `adduser` implementations actually write. Moving the path costs one line.

Checked against the file each system generates:

| | before | after |
|---|---|---|
| OpenBSD, as written by `adduser(8)` | `parse_failed` | 17 settings |
| Debian 3.152, every shipped setting uncommented | 31 settings | identical tree |
| FreeBSD, as written by `save_config()` | 11 settings | identical tree |

Saving with no changes leaves all three byte-identical, and setting one key
rewrites only that line — including the perl list:

```
$ augtool --autosave -e "set /files/etc/adduser.conf/shellpref \"('sh', 'ksh')\""
```

What is lost is shell syntax beyond plain assignment. Of the forms plausible in
this file only `export` differs: `defaultgroups="wheel operator"` and
`homeprefix=/home/$company` parse identically under both lenses. `export` does
not change how FreeBSD's `adduser` uses the file, since it sources it into its
own shell.

`augparse` passes on all 232 lens test modules.

OpenBSD's ports tree has carried a local patch dropping the path from
`Shellvars` since 2016, noting *"adduser.conf is not a shell script, so don't
try to parse it as such"*. Dropping it outright would lose the file on Debian
and FreeBSD, which is presumably why it stayed local.

---

The Build check will come back red: `.github/workflows/build.yml` still asks
for `ubuntu-20.04`, and every run since January has waited a day for a runner
and then been cancelled. It is not this change.


georgehansper:
Thank you vey much for the detailed analysis, and the real-world use-cases mentioned

This is a good idea

georgehansper (commented):
Thank you for the extra tests

georgehansper (approved):

PR #898, opened 2026-09-19
Get the CI and release workflows running again

lutter:
[ Forgive the AI slop in the description, it's Saturday ]

Neither of the two workflows in this repo could run. This gets them working again, and removes the leftovers of the CI that preceded them.

### The build workflow never started

It pinned `runs-on: ubuntu-20.04`, a runner image GitHub has retired, so every run failed before executing a step, with only *"This run likely failed because of a workflow file issue"* to go on. Every open pull request therefore carries a red X
that says nothing about its contents.

A second, independent breakage sat behind it: the `.gnulib` submodule URL was `git://git.sv.gnu.org/gnulib.git`, and savannah no longer serves the unauthenticated `git://` protocol, so `bootstrap` cannot fetch gnulib on a fresh
clone.

Beyond repairing those:

* Run on `ubuntu-latest` rather than a dated label, so the workflow does not rot  the same way again
* Configure with `--disable-gnulib-tests`. That drops gnulib's own test suite from  `SUBDIRS`; it tests gnulib against the local libc rather than testing augeas, and  was the reason the workflow patched `gnulib/tests/test-readlink.h` by line number. That `sed -i '41s/ENOENT/.../'` step is gone
* Build with gcc and with clang. Both are green; clang additionally reports 69  warnings that gcc does not, notably 10
  `-Wsingle-bit-bitfield-constant-conversion`
* Install the dependencies the tree actually needs (bison, flex, libselinux1-dev)  rather than relying on whatever the runner image ships
* Print `tests/test-suite.log` when the suite fails, since the log is not otherwise  part of the run
* Update `actions/checkout` off `@master`, build in parallel, and cancel superseded runs of the same ref

### The release workflow could not have produced a release

It used `::set-output`, which GitHub stopped honoring in 2022, and `actions/create-release` and `actions/upload-release-asset`, both of which GitHub has archived and no longer runs. It also installed only git and build-essential, so
`make distcheck` had neither bison nor flex nor libxml2 to configure against.

* Create the release with `gh release create` rather than the archived actions,  which also removes two third-party actions from the release path. It still  creates a draft
* Report the version through `$GITHUB_OUTPUT`
* Check out the full history: the `dist` target generates ChangeLog with  `gitlog-to-changelog`, and with the default shallow checkout that ChangeLog would  hold a single entry
* Fail the release if the tag does not match `AC_INIT` in `configure.ac`, rather  than looking for a tarball whose name does not match the tag
* Set `DISTCHECK_CONFIGURE_FLAGS` in the top `Makefile.am`, so that `make distcheck`  works for everyone and not just in CI

### Leftovers

* `tests/Makefile.am` used the obsolete `INCLUDES` rather than `AM_CPPFLAGS`.  automake still expands it, so nothing was broken; this only removes the warning  before automake drops the alias. No target there sets its own `_CPPFLAGS`, which would not inherit `AM_CPPFLAGS` the way it inherits `INCLUDES`, so the include  path is unchanged
* The README build badge pointed at travis-ci.org. It did not render as broken:  travis still serves a cached SVG reading "build | passing" in green, so the README  has been advertising a green build for a CI system that last ran in 2021. It now points at the Actions workflow. The fuzzing badge is still served and still  accurate; only its link needed updating, since OSS-Fuzz moved its issues off  bugs.chromium.org
* `.travis.yml` is removed. CI moved to Actions in #714 in January 2021 and the file  has not been touched since October of that year; travis-ci.org is shut down, the  project never moved to travis-ci.com, its IRC notifications point at freenode, and its build recipe is a subset of `build.yml`

### Testing

Verified locally on Fedora 43 (gcc 15.3, clang, automake 1.18):

| | build | `make check` |
|---|---|---|
| gcc | exit 0 | 260 PASS / 4 SKIP / 0 FAIL |
| clang | exit 0 | 260 PASS / 4 SKIP / 0 FAIL |

`make distcheck` passes as well (exit 0, same totals, tarball produced), which is what exercises the `AM_CPPFLAGS` change, since in-tree `top_srcdir` and `top_builddir` coincide and cannot.

Two things could not be verified locally: `./src/try valgrind` (valgrind is not installed here) and the tag-triggered release path, which needs a `release-*` tag pushed. The build badge will read "failing" until this merges and the first run on
master succeeds.

The first commit is cherry-picked from the in-flight oss-fuzz branch, since without it CI cannot fetch gnulib at all; it will deduplicate when that branch lands.


georgehansper (approved):
Looks good, thanks

georgehansper:
I only looked at the github actions changes superficially, so you might want a second opinion on those
The other changes were more straightforward, and are perfectly fine

PR #899, opened 2026-09-20
Fix aug_preview() when span is enabled (was failing with "Lens not found")

georgehansper:
This is an alternative solution to Pull Request #859

This solution updates aug_preview() and does make any changes to tree_source()


src/augeas.c
```diff
@@ -2000,11 +2000,11 @@ int aug_preview(struct augeas *aug, const char *path, char **out) {
     ERR_BAIL(aug);
     ERR_THROW(tree == NULL, aug, AUG_ENOMATCH, "No node matching %s", path);
 
-    file_path = tree_source(aug, tree);
-
-    ERR_THROW(file_path == NULL, aug, AUG_EBADARG, "Path %s is not associated with a file", path);
+    while (!(ROOT_P(tree) || tree->file))
```
lutter, line 2003: I had to scratch my head on this condition a little. It might be clearer to say `while (!ROOT_P(tree) && tree->file == NULL) { .. }`

lutter (approved):

